PR #1, opened 2021-11-17
Initial Commit

NicholasFan235:
Initial cpp framework for epiabm simulation


src/dataclasses/types.hpp
```diff
@@ -0,0 +1,42 @@
...
+        Susceptible,
+        Exposed,
+        InfectASympt,
+        InfectMild,
+        InfectGP,
+        InfectHosp,
+        InfectICU,
+        InfectICURecov,
+        Recovered,
+        Dead
```
rcw5890, line 36: Supposing we want to use the software to do a comparison of alternative within-host dynamics which use different categories---would we just have to declare another enum? How will this work with m_status in the Person class?

rcw5890 (changes requested):
Partial review
I'm learning C++ as I go so please excuse novice questions
Overall everything is installing nicely on my system, and I think the architecture will be fantastically useful---many thanks

build/test.py
```diff
@@ -0,0 +1,22 @@
+import covidsim as cs
+
+f = cs.Factory()
+population = f.make_population()
+
+f.add_cells(population, 5)
```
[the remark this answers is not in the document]
NicholasFan235: Agree, having it in the factory is a little dodgy.

Came down to wanting to make it hard to add or remove cells, microcells and people mid simulation running to avoid having to do extra checks about the size of the vectors at each iteration.

I think these methods are going to be made private to be used as population helper functions, and this PopulationFactory class is instead going to just expose make_population(n_cells, n_microcellsPerCell, n_peoplePerMicrocell) which will construct a basic population and return it.

Maybe in future we can also create a PopulationFactoryFromData class which will read in a datafile and construct the population from that.

src/dataclasses/types.hpp
NicholasFan235: Right now, would have to modify this enum, but it would be easy to modify to allow passing in different enums. (Probably using templating)

KCGallagher (approved):
We talked through this earlier and it all looks good! Hopefully we can merge the docs branch into this before merging to main, but I know that relies on the docs compiling first. 

Also maybe worth having some installation/running advice in the readme, targetted at ubuntu developers (i.e. us) just to help those (like myself) less familiar with cmake etc - I know it should be very simple but just where to run the code / how to specify the build directory etc.

Thanks so much for all your work on this!

rcw5890:
> We talked through this earlier and it all looks good! Hopefully we can merge the docs branch into this before merging to main, but I know that relies on the docs compiling first.
> 
> Also maybe worth having some installation/running advice in the readme, targetted at ubuntu developers (i.e. us) just to help those (like myself) less familiar with cmake etc - I know it should be very simple but just where to run the code / how to specify the build directory etc.
> 
> Thanks so much for all your work on this!

Agreed, I made a few notes on this while working through the compilation myself. We should expand it later but it covers the basic procedure.

cEpiabm/src/dataclasses/person.hpp
```diff
@@ -0,0 +1,66 @@
...
+        Susceptible,
+        Exposed,
+        InfectASympt,
+        InfectMild,
+        InfectGP,
+        InfectHosp,
+        InfectICU,
+        InfectICURecov,
+        Recovered,
+        Dead
```
I-Bouros, line 27: The E, R and D currently miss from the python implementation. Should we also add them there?

I-Bouros (changes requested):
I think overall it's great. Minor comments. I think though more comments are needed overall to make it more understandable (for us as well) about what each step is doing (and more importantly what each of the quantities there is aiming to achieve).

rcw5890 (approved):

I-Bouros (approved):

PR #5, opened 2021-11-24
Pyinit

KCGallagher:
python backend package


KCGallagher:
Will add CI before the pull request by 25/11

KCGallagher:
N.B codecov badge isn't currently happy, but I believe that's because the badge is based on the coverage in the main branch, which is currently undefined. Shouldn't be an issue post merge, but I promise to fix it if it is!

KCGallagher:
@rcw5890 Got almost all the tests working - just having issues with the docs test - its giving error 2 which I can't find in the code. 

Its trying to find the source directory at `/home/runner/work/epiabm/epiabm/docs/source`, which I assume should be `/home/runner/work/epiabm/pyEpiabm/docs/source` but can't find where in the code this comes from.

It's possible it relates to the comments on [this](https://stackoverflow.com/questions/53668052/sphinx-cannot-find-my-python-files-says-no-module-named) question, which suggests changing the `conf.py` file to reflect the new directory structure (as we effectively have an extra folder compared to the seirmo structure), however I am not sure how to go about this!

If you could have a look at this, that would be very much appreciated!

KCGallagher:
Also, I have not implemented branch protection status checks, as I believe these would apply to the C++ code too, and I'm not sure if there's a way to work around this?

I-Bouros (approved):
Just as a point: Would we want the documentation for the two versions pyEpiabm and cppEpiabm in the same place? In that case it might make more sense to have the ``docs`` at the topmost level?

rcw5890:
Thanks for your work on this @KCGallagher 
As you suggest, the doctest problem was caused by the directory path. However, the easiest way to fix this is just to do a cd in the workflow, so we don't have to change any of the other files.

Additionally, sphinx was picky about your plural classes because it wants a space after the `. If you don't mind unicode docstrings, you can use the thin spaces (see the latest population.py). 

The branch protection rules will apply to all pull requests including C++ ones, but that's fine as a C++ pull request should not cause any of the python checks to fail (and if it does, we'd want to be aware). We can exclude C++ directories as necessary from the python checks if they cause problems.

rcw5890:
> Just as a point: Would we want the documentation for the two versions pyEpiabm and cppEpiabm in the same place? In that case it might make more sense to have the `docs` at the topmost level?

Possibly, but since we might not even be using sphinx for c++ (I don't know if it's the best choice for that language) I would worry about this later. 
Certainly we should make it easy for the user to get from one documentation to the other.

pyEpiabm/docs/source/core.rst
```diff
@@ -0,0 +1,29 @@
+****
+Core
+****
+
+.. currentmodule:: pyEpiabm
+
+Core provides a framework for the implementation of SEIR and Related Models.
```
rcw5890, line 7:
```suggestion
...
```

pyEpiabm/pyEpiabm/cell.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Cell:
+    """Class representing a Cell (Subset of Population)
+    Collection of :class:`Microcell` s and :class:`Person` s
+    """
+    def __init__(self):
+        """Constructor Method
+        """
+        self.microcells = []
+        self.people = []
```
rcw5890, line 12:
```suggestion
        self.persons = []
```
and so forth. Not important but this would be slightly more consistent

pyEpiabm/setup.py
```diff
@@ -0,0 +1,84 @@
+
+#
+# epiabm setuptools script
```
rcw5890, line 3:
```suggestion
# pyepiabm setuptools script
```

pyEpiabm/setup.py
```diff
@@ -0,0 +1,84 @@
...
+#
+# This file is part of ePIABM
```
rcw5890, line 5:
```suggestion
# This file is part of epiabm
```

pyEpiabm/setup.py
```diff
@@ -0,0 +1,84 @@
...
+# (https://github.com/SABS-R3-Epidemiology/epiabm.git) which is released
+# under the BSD 3-clause license. See accompanying LICENSE.md for copyright
+# notice and full license details.
```
rcw5890, line 8: As it stands there is no license.md, so the rest of this comment can be removed

pyEpiabm/setup.py
```diff
@@ -0,0 +1,84 @@
...
+from setuptools import setup, find_packages
+
+
+def get_version():
+    """
+    Get version number from the pyEpiabm module.
+    The easiest way would be to just ``import pyEpiabm ``, but note that this may  # noqa
+    fail if the dependencies have not been installed yet. Instead, we've put
+    the version number in a simple version_info module, that we'll import here
+    by temporarily adding the oxrse directory to the pythonpath using sys.path.
```
rcw5890, line 19:
```suggestion
    The easiest way would be to just ``import pyEpiabm ``, but note that this
    may fail if the dependencies have not been installed yet. Instead, we've
    put the version number in a simple version_info module, that we'll import
    here by temporarily adding the oxrse directory to the pythonpath using
    sys.path.
```

pyEpiabm/setup.py
```diff
@@ -0,0 +1,84 @@
...
+def get_readme():
+    """
+    Load README.md text for use as description.
+    """
+    with open('README.md') as f:
+        return f.read()
+
+
+setup(
+    # Module name (lowercase)
```
rcw5890, line 40:
```suggestion
    # Package name
```

rcw5890 (changes requested):

pyEpiabm/docs/source/index.rst
```diff
@@ -0,0 +1,25 @@
+.. pyEpiabm documentation master file, created by
+   sphinx-quickstart on Wed Oct 28 05:13:22 2020.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to pyEpiabm's documentation!
+====================================
+
+This is a python backend for the EpiAbm package
```
I-Bouros, line 9:
```suggestion
This is a python backend for the EpiAbm package.
```

pyEpiabm/pyEpiabm/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .person import Person  # noqa
```
I-Bouros, line 1: Add some heading at the top to explain the purpose of the file. Try following the ``branchpro`` former with the licensing ad stuff?

pyEpiabm/pyEpiabm/cell.py
```diff
@@ -0,0 +1,30 @@
+from .microcell import Microcell
```
I-Bouros, line 1: Same comment as before for all the classes files.

pyEpiabm/pyEpiabm/cell.py
```diff
@@ -0,0 +1,30 @@
...
+        return f"Cell with {len(self.microcells)} microcells " + \
+            f"and {len(self.people)} people."
+
+    def add_microcells(self, n):
+        """Add n empty :class:`Microcell` s to Cell
+
+        :param n: Number of empty :class:`Microcell` s to add
+        :type n: int
+        """
+        for i in range(n):
```
I-Bouros, line 29:
```suggestion
        for _ in range(n):
```

Same for all the other files

pyEpiabm/pyEpiabm/cell.py
```diff
@@ -0,0 +1,30 @@
...
+        """String representation of Cell
+
+        :return: String representation of Cell
+        :rtype: str
+        """
+        return f"Cell with {len(self.microcells)} microcells " + \
+            f"and {len(self.people)} people."
+
+    def add_microcells(self, n):
+        """Add n empty :class:`Microcell` s to Cell
```
I-Bouros, line 24:
```suggestion
        """Add n empty :class:`Microcell` s to Cell.
```

pyEpiabm/pyEpiabm/cell.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self):
+        """Constructor Method
+        """
+        self.microcells = []
+        self.people = []
+
+    def __repr__(self):
+        """String representation of Cell
+
+        :return: String representation of Cell
```
I-Bouros, line 17:
```suggestion
        :return: String representation of Cell.
```

pyEpiabm/pyEpiabm/cell.py
```diff
@@ -0,0 +1,30 @@
+from .microcell import Microcell
+
+
+class Cell:
+    """Class representing a Cell (Subset of Population)
+    Collection of :class:`Microcell` s and :class:`Person` s
```
I-Bouros, line 6:
```suggestion
    Collection of :class:`Microcell` s and :class:`Person` s.
```

pyEpiabm/pyEpiabm/cell.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    def __init__(self):
+        """Constructor Method
```
I-Bouros, line 9: Not sure why we should need a doctoring for the ``__init__`` method:)), but don't forget the `.` at the end of each docstring

pyEpiabm/pyEpiabm/microcell.py
```diff
@@ -0,0 +1,37 @@
+from .person import Person
```
I-Bouros, line 1: Header for the test files

pyEpiabm/pyEpiabm/version_info.py
```diff
@@ -0,0 +1,4 @@
+VERSION_INT = 0, 0, 1
```
I-Bouros, line 1: + header

I-Bouros (changes requested):
It's great! Minor comments about styling more than anything else. Also to help with the test command you can fave that ``__init__.py`` file for the tests folder as we have in ``branchpro``? Up to you:))

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,35 @@
...
+    branches:
+      - '**'
+
+jobs:
+
+  build-and-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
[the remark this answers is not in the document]
KCGallagher: Seems sensible! - Sorry this was just copied from the older branchpro code when it probably wasn't relevant! I suppose it's possible that the testing could become expensive as the code base grows, but right now there won't be any issues.

pyEpiabm/pyEpiabm/cell.py
KCGallagher: I think PEP8 says that all public methods (including __init__()) should have docstrings - will make sure they all have full stops at the end though!

pyEpiabm/pyEpiabm/cell.py
KCGallagher: Okay so 'persons' is the list of individuals, and 'people' is the length of that list effectively (following your OED definition). I made this change last (in a separate commit) so it can be easily checked/reverted if you don't like it.

I-Bouros (approved):

pyEpiabm/docs/source/core.rst
```diff
@@ -0,0 +1,29 @@
+****
+Core
+****
+
+.. currentmodule:: pyEpiabm
+
+...
```
rcw5890, line 7: Looks good, thanks for making all the changes

But by the "..." I meant replace that seirmo sentence with an appropriate one for epiabm !

pyEpiabm/run_tests.py
```diff
@@ -0,0 +1,269 @@
...
+import re
+import subprocess
+import unittest
+
+
+def run_unit_tests():
+    """
+    Runs unit tests (without subprocesses).
+    """
+    tests = os.path.join('pyEpiabm', 'pyEpiabm', 'tests')
```
rcw5890, line 20:
```suggestion
    tests = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'pyEpiabm', 'tests')
```
Currently, it's oversensitive to where the run_test.py is called from---this should fix that

rcw5890 (changes requested):

pyEpiabm/run_tests.py
KCGallagher: Ah brilliant - this was frustrating me!

pyEpiabm/docs/source/core.rst
KCGallagher: My bad -  I thought this was a syntax thing! Now corrected

rcw5890 (approved):

PR #12, opened 2021-12-01
Test sweep

Elizabeth-Hayman:


KCGallagher:
Main method to enforce coverage requirements seems to be through a codecov.yml file, which has a different format to our coverage-test.yml file and I'm not quite sure how to integrate them (having both seems to do nothing).
Either may, we may want to sort this before merging, or certainly increase coverage to 100%!

pyEpiabm/pyEpiabm/tests/test_household.py
```diff
@@ -0,0 +1,18 @@
...
+class TestHousehold(unittest.TestCase):
+    """
+    Test the 'Household' class.
+    """
+    def test_construct(self):
+        pe.Household((1, 1))
+
+    def test___repr__(self):
+        subject = pe.Household((1, 1))
+        self.assertIsInstance(repr(subject), str)
```
KCGallagher, line 14: Might be good to test whether the actual string is as you expect too, and it's not just printing a null case?

pyEpiabm/pyEpiabm/tests/test_household_sweep.py
```diff
@@ -0,0 +1,76 @@
+import unittest
+import pyEpiabm as pe
+from queue import Queue
+from pyEpiabm.household_sweep import HouseholdSweep
+from unittest import mock
+
+
+class TestHouseholdSweep(unittest.TestCase):
+    """
+    Test the 'HouseholdSweep' class.
```
KCGallagher, line 10:
```suggestion
    """Test the 'HouseholdSweep' class.
```

pyEpiabm/pyEpiabm/tests/test_household_sweep.py
```diff
@@ -0,0 +1,76 @@
...
+        self.person.infection_status = pe.InfectionStatus.InfectMild
+        test_queue = Queue()
+        new_person = pe.Person(self.microcell)
+        new_person.household = self.house
+        self.house.persons.append(new_person)
+        self.pop.cells[0].persons.append(new_person)
+
+        test_queue.put(new_person)
+        subject(time, self.pop)
+        self.assertEqual(self.cell.person_queue.qsize(), 1)
```
KCGallagher, line 64: Am I misunderstanding - you said that both people should be added to the queue but then test for a queue length of 1, why is this?

KCGallagher (changes requested):
Overall like this, just some minor consistency things in the docstrings mainly! I don't understand the role of the parameters.py file so have skipped this

pyEpiabm/pyEpiabm/household_sweep.py
```diff
@@ -0,0 +1,56 @@
...
+
+class HouseholdSweep(AbstractSweep):
+    '''Class to run the intra-household infections
+    as part of the sweep function. Takes an individual
+    person as input and tests a infection event against each
+    susceptible member of their household. The resulting
+    exposed person is added to an infection queue.
+
+    : param infected: Person
+    : return
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Hmm not sure what the convention is when nothing is returned, just an attribute of a different object changed?

pyEpiabm/pyEpiabm/household_sweep.py
```diff
@@ -0,0 +1,56 @@
...
+    '''
+
+    def __call__(self, time: float, population: Population):
+        '''
+        Given a population structure, loops over infected members
+        and considers whether they infected household members based
+        on individual, and spatial infectiousness and susceptibility.
+        '''
+        timestep = int(time * Parameters.instance().time_steps_per_day)
```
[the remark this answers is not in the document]
Elizabeth-Hayman: I don't really know, its something related to the parameters file, better check with Nick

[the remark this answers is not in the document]
KCGallagher: @NicholasFan235 can you adivse?

pyEpiabm/pyEpiabm/covidsim_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+        :type infectee: Person
+        :param timestep: Current simulation timestep
+        :type timestep: int
+        :return: susceptibility
+        :rtype: float
+        """
+        return CovidsimHelpers.calc_person_susc(infector, infectee, timestep)
+
+    @staticmethod
+    def calc_person_susc(infector: Person, infectee: Person, timestep: int):
```
KCGallagher, line 42: I know we haven't implemented them yet, but do we anticipate that these functions will have explicit time dependence (and that an individual's susceptibility depends on who is infecting them - not clear if this is an interaction property or a property of an individual? If it is a property of an individual, do they have one for each possible infector?)

Not a big issue but maybe worth being consistent now for what we plan to add later

pyEpiabm/pyEpiabm/household.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __repr__(self):
+        """String representation.
+
+        :return: String representation of the household
+        :rtype: str
+        """
+        return "Household at " \
+            + f"({self.location[0]:.2f}, {self.location[1]:.2f}) "\
+            + f"with {len(self.persons)} persons."
```
KCGallagher, line 35:
```suggestion
            + f"with {len(self.persons)} people."
```

(As we are only counting them here - to be consistent with `cell.py` etc?

KCGallagher: (Of course, change the testing file too if you change this)

pyEpiabm/pyEpiabm/tests/test_household_sweep.py
```diff
@@ -0,0 +1,74 @@
...
+        adds persons to the queue.
+        '''
+        mock_inf.return_value = 10
+        mock_susc.return_value = 10
+        subject = HouseholdSweep()
+        time = 1
+
+        # Assert a population with one infected will not change the queue
+        subject(time, self.pop)
+        assert(self.cell.person_queue.empty())
```
KCGallagher, line 44: Should we be using self.assertTrue here?

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/household.py
Elizabeth-Hayman: Should be consistently persons now i think

pyEpiabm/pyEpiabm/tests/test_household_sweep.py
Elizabeth-Hayman: Probably

pyEpiabm/pyEpiabm/covidsim_helpers.py
Elizabeth-Hayman: Say we have an immune resistance decrease fucntion which is pulled in within the inf/susc function and needs that time dependence

KCGallagher: Ooh okay yep that makes sense, and we could add in some internal variable to store infection time to convert absolute simulation time to relative infection time.

Still not sure about the infector argument though, I thought susceptibility was a property of an individual not a potential infection opportunity, so why could it depend on the infector object? Think it should only take one person object as an argument (but happy to be convinced otherwise!)

pyEpiabm/pyEpiabm/household_sweep.py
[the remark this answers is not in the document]
KCGallagher: @rcw5890 I wanted to ask you about the conventions here anyway, because we've got a few going on in different parts of the code, and it's good to be consistent (and compatible with the docs)

pyEpiabm/pyEpiabm/household_sweep.py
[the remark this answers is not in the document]
NicholasFan235: Covidsim does it this way (iterates timesteps, converts to time in days, but within many sweeps converts back to timesteps), not sure I like it.

Parameters uses a 'Singleton' pattern so it can be accessed everywhere without being passed around. It's sort of a 'better practice global'

pyEpiabm/pyEpiabm/household.py
```diff
@@ -0,0 +1,35 @@
...
+    have a combined susceptability and infectiousness
+    different to that of the individuals.
+    """
+    def __init__(self, loc: typing.Tuple[float, float],
+                 susceptibility=0, infectiveness=0):
+        """Constructor Method.
+
+        :param loc: Location of household.
+        :type loc: Tuple[float, float]
+        :param susceptibility: Household's
```
KCGallagher, line 20: This docstring is incomplete as it doesn't cover all the parameters

pyEpiabm/pyEpiabm/person.py
```diff
@@ -29,6 +29,28 @@ def __init__(self, microcell,
...
+            InfectionStatus.InfectGP,
+            InfectionStatus.InfectHosp,
+            InfectionStatus.InfectICU,
+            InfectionStatus.InfectICURecov]
+
+    def is_susceptible(self):
+        """Query if the person is currently susceptible.
+
+        :return: Whether person is currently suscptible
+        :rtype: bool"""
```
KCGallagher, line 52:
```suggestion
        :rtype: bool
        """
```

KCGallagher (changes requested):

KCGallagher (approved):

pyEpiabm/pyEpiabm/tests/test_parameters.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class TestParameters(unittest.TestCase):
+    """Test the 'Parameters' class.
+    """
+    def test_instance(self):
+        pe.Parameters.instance()
+
+    def test_setattr(self):
+        pe.Parameters.instance().abc = 1
```
I-Bouros, line 12: what is the actual test? You can maybe collapse the setattr and getattr into one instead

I-Bouros (changes requested):

pyEpiabm/pyEpiabm/person.py
```diff
@@ -40,9 +40,32 @@ def __init__(self, microcell,
...
+            InfectionStatus.InfectMild,
+            InfectionStatus.InfectGP,
+            InfectionStatus.InfectHosp,
+            InfectionStatus.InfectICU,
+            InfectionStatus.InfectICURecov]
+
+    def is_susceptible(self):
+        """Query if the person is currently susceptible.
+
+        :return: Whether person is currently suscptible
```
rcw5890, line 64:
```suggestion
        :return: Whether person is currently susceptible
```

rcw5890 (changes requested):

pyEpiabm/pyEpiabm/tests/test_parameters.py
KCGallagher: @NicholasFan235 can you advise on this - I never really understood this file!

pyEpiabm/pyEpiabm/__init__.py
```diff
@@ -9,10 +9,17 @@
...
 from .person import Person  # noqa
 from .population import Population  # noqa
 from .cell import Cell  # noqa
 from .microcell import Microcell  # noqa
-from .infection_status import InfectionStatus # noqa
-from .abstract_sweep import AbstractSweep # noqa
+from .household import Household  # noqa
+from .infection_status import InfectionStatus  # noqa
+from .parameters import Parameters  # noqa
+from .covidsim_helpers import CovidsimHelpers  # noqa
```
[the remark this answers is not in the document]
KCGallagher: That is very nice!

pyEpiabm/pyEpiabm/tests/test_parameters.py
NicholasFan235: These tests mostly just for code coverage completeness.
Parameters is setup as a singleton, i.e. 'Better practice version of a global variable'
Essentially you setup a global variable, but it is managed by a class so you don't have to worry about whether it has been created.
These tests are just checking that you can read and write to the global Parameters object. (mostly for coverage completeness)

rcw5890 (approved):

pyEpiabm/pyEpiabm/tests/test_parameters.py
KCGallagher: The `setattr` unit test doesn't affect coverage and is effectively repeated in the `getattr` unit test, so I agree they should probably be combined and have changed that

pyEpiabm/pyEpiabm/parameters.py
```diff
@@ -0,0 +1,25 @@
...
+    class __Parameters:
+        """Singleton Parameters Object
+        """
+        def __init__(self):
+            pass
+
+    _instance = None  # Singleton instance
+
+    @staticmethod
+    def instance():
```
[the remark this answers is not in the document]
KCGallagher: I have updated the documentation a little bit so hopefully this is clearer without being too verbose!

I-Bouros (approved):

PR #27, opened 2021-12-14
C++ Logging Classes

NicholasFan235:
Logging functionality
Created using Singleton Logger class
Multithreading framework setup
Output folder handling class


KCGallagher (approved):
I think these look great! I was wondering whether we should document the roles of each logging level somewhere, or are these standard levels so that competent users would know what to expect at each level?

rcw5890 (changes requested):

rcw5890 (approved):

PR #36, opened 2021-12-16
Working Simulation in C++

NicholasFan235:
Added Simulation classes and bunch of other helpers.
Modifications across many classes.
Modified Place to allow linking people in different cells
Added reporter classes which monitor the simulation and output data to files.


cEpiabm/src/reporters/percell_compartment_reporter.cpp
```diff
@@ -0,0 +1,88 @@
...
+            InfectionStatus::Susceptible,
+            InfectionStatus::Exposed,
+            InfectionStatus::InfectASympt,
+            InfectionStatus::InfectMild,
+            InfectionStatus::InfectGP,
+            InfectionStatus::InfectHosp,
+            InfectionStatus::InfectICU,
+            InfectionStatus::InfectICURecov,
+            InfectionStatus::Dead,
+            InfectionStatus::Recovered
```
KCGallagher, line 23: I don't like that we have to define these in the `cell_compartment_reporter` and the `percell_compartment_reporter` - is there any way to inherit/import these from a single place (so that we would only have to change them in one place if we wanted to change the list of compartments

cEpiabm/src/simulations/basic_simulation.hpp
```diff
@@ -0,0 +1,47 @@
...
+    };
+
+    typedef std::shared_ptr<BasicSimulation> BasicSimulationPtr;
+
+}
+
+
+
+
+#endif // EPIABM_SIMULATINS_BASIC_SIMULATION_HPP
```
KCGallagher, line 47:
```suggestion


#endif // EPIABM_SIMULATINS_BASIC_SIMULATION_HPP

```

Ditto on empty lines

cEpiabm/src/sweeps/household_sweep.cpp
```diff
@@ -78,7 +78,7 @@ namespace epiabm
 
         double infectiousness = Covidsim::CalcHouseInf(infector, timestep);
         double susceptibility = Covidsim::CalcHouseSusc(infector, infectee, timestep);
-        double foi = infectiousness * susceptibility;
+        double foi = infectiousness * susceptibility * 0.5;
 
         if (static_cast<double>(std::rand() % 1000000) / static_cast<double>(1000000) < foi)
```
KCGallagher, line 83: I am confused about the factor of 1/2 here - this isn't currently consistent with pyEpiabm but should we add it there?

And why the use of a million in L83?

cEpiabm/test/dataclasses/test_person.cpp
```diff
@@ -40,3 +46,29 @@ TEST_CASE("dataclasses/person: test setHousehold", "[Person]")
         static_cast<size_t>((std::rand()%500) * 2 + 1)) == false);
```
KCGallagher, line 46: Does that rely on it producing the same random number as the previous call?

cEpiabm/test/dataclasses/test_place.cpp
```diff
@@ -1,25 +1,149 @@
...
+        subject.cells().push_back(Cell(i));
+        subject.cells()[i].people().reserve(n_people);
+        for (size_t j = 0; j < n_people; j++)
+        {
+            subject.cells()[i].people().push_back(Person(i, j, j));
+
+            /*printf("Microcell: %ld, Person: (%ld, %ld)\n",
+                i,
+                subject.people()[subject.microcells()[i].people()[j]].cellPos(),
+                subject.people()[subject.microcells()[i].people()[j]].microcellPos());//*/
```
KCGallagher, line 32:
```suggestion
```

Fine for testing but I think commented out code is generally not a good thing?

KCGallagher (commented):
Looks really good - I have added minor comments but mostly on things I don't understand - apologies for the stupid questions!

P.S There is an empty file `simulation_interface.hpp` - I can't comment on it directly because it has no content but maybe this should be deleted?
P.P.S When you comment `namespace epiabm` at the final closing bracket, is that just good practice to document the extent of code using this namespace?

cEpiabm/examples/basic_example/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "sweeps/new_infection_sweep.hpp"
+#include "sweeps/basic_host_progression_sweep.hpp"
+#include "sweeps/random_seed_sweep.hpp"
+#include "reporters/cell_compartment_reporter.hpp"
+#include "reporters/percell_compartment_reporter.hpp"
+#include "reporters/population_compartment_reporter.hpp"
+
+#include <random>
+
+void run()
```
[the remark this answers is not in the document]
NicholasFan235: More of a testing script, in the future probably going to expose it to python. Could comment here, or in a jupyter notebook might be better

cEpiabm/examples/basic_example/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+    using namespace epiabm;
+
+    std::srand(100); // random seed
```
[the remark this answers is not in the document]
NicholasFan235: Right now it sets it for everything, but this probably should be changed in the future. Currently all the random number generation is just from uniform distributions, need to create some random number classes probably.

cEpiabm/examples/basic_example/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+    LogFile::Instance()->configure(2, std::filesystem::path("output/log.log"));
+
+    // Make Population and Link Households
+    PopulationPtr population = PopulationFactory().makePopulation(10, 10, 1000);
+    HouseholdLinker().linkHouseholds(population, 1, 100);
+    population->initialize();
+
+    // Randomly Seed Population with Infections
+    {
+        RandomSeedSweep randomizer = RandomSeedSweep(600);
```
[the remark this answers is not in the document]
NicholasFan235: I'm not sure named parameters are in the C++ standard. I think some compilers are okay with it depending on parameters. Seems like GCC supports it from C++20, but MSVC you need to add extra flags. You'd normally add those as a comment

cEpiabm/src/reporters/percell_compartment_reporter.cpp
NicholasFan235: Could create a global variable which stores a list of all statuses.

cEpiabm/src/sweeps/basic_host_progression_sweep.cpp
```diff
@@ -95,7 +95,7 @@ namespace epiabm
      */
     inline unsigned short BasicHostProgressionSweep::time_to_next_status(Person* /*person*/, InfectionStatus /*status*/)
     {
-        return static_cast<unsigned short>(std::rand() % 10 + 1);
+        return static_cast<unsigned short>(std::rand() % 14 + 14);
```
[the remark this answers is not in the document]
NicholasFan235: Just an arbitrary random number for now. I don't think it will be a uniform distribution in the future, so parameters don't really matter

cEpiabm/src/sweeps/household_sweep.cpp
NicholasFan235: I think covid-sim had that factor, could be wrong. Used million to generate random float to 6dp, in the future I imagine this will be handled by some random number classes elsewhere

cEpiabm/test/dataclasses/test_person.cpp
NicholasFan235: No, checking setHousehold returns false if the person is already linked to a household. (easier not to allow changing household)

cEpiabm/test/dataclasses/test_place.cpp
NicholasFan235: Fine to have in the tests, this bit of code is pretty tricky, and its possible we'd have to modify it in the future so it'll be useful to have these prints handy. Very easy to mess up initialization and get segmentation faults with this function.

cEpiabm/src/reporters/percell_compartment_reporter.cpp
KCGallagher: Yeah that definitely seems tidier and less error-prone/easier for the future if that's easy to do!

cEpiabm/src/sweeps/household_sweep.cpp
KCGallagher: Ooh okay - will have a look as the python doesn't have this so one backend needs to be changed!

cEpiabm/test/dataclasses/test_place.cpp
KCGallagher: Sorry I didn't notice this was a test file - all good!

cEpiabm/src/sweeps/household_sweep.cpp
KCGallagher: @NicholasFan235 have been going through the CovidSim code on this and I really can't find this factor of 1/2 - have you got a reference for it?

cEpiabm/src/covidsim.cpp
```diff
@@ -25,6 +41,14 @@ namespace epiabm
...
+    double Covidsim::CalcSpaceSusc(
+        Cell* /*cell*/,
+        Person* /*infectee*/,
+        unsigned short int )
+    {
+        return 0.2;
+    }
 
 
 } // namespace epiabm
```
KCGallagher, line 54:
```suggestion
} // namespace epiabm

```
Seems to want newline characters at the end of files

cEpiabm/src/dataclasses/infection_status.hpp
```diff
@@ -0,0 +1,42 @@
...
+            case InfectionStatus::InfectICU: return "InfectICU";
+            case InfectionStatus::InfectICURecov: return "InfectICURecov";
+            case InfectionStatus::Recovered: return "Recovered";
+            case InfectionStatus::Dead: return "Dead";
+            default: return "Unknown";
+        }
+    }
+}
+
+#endif // EPIABM_DATACLASSES_INFECTION_STATUS_HPP
```
KCGallagher, line 42:
```suggestion
#endif // EPIABM_DATACLASSES_INFECTION_STATUS_HPP

```
Another warning for missing end-line character - not wanting to self publicise but I think setting up the `.editorconfig` file could avoid these?

cEpiabm/src/dataclasses/microcell.cpp
```diff
@@ -1,27 +1,30 @@
...
         m_households(),
-        m_cellPos(cellPos)
+        m_cellPos(cellPos),
+        m_compartmentCounter()
     {}
 
+    Microcell::~Microcell()
+    {
+        //std::cout << "Microcell Destructor" << std::endl;
+    }
```
KCGallagher, line 18: Do we still need this (I know you have it in a few files, like `person.cpp` as well)?

cEpiabm/src/dataclasses/person.hpp
```diff
@@ -41,18 +34,20 @@ namespace epiabm
...
+        std::set<size_t> m_places; // Indices of Places which the person is a member of;
 
     public:
         Person(size_t microcell, size_t cellPos, size_t mcellPos);
-        ~Person() = default;
-        Person(const Person&) = default;
-        Person(Person&&) = default;
+        ~Person();
+        //Person(const Person&) = default;
+        //Person(Person&&) = default;
```
KCGallagher, line 45: Should these still be commented?

cEpiabm/src/population_factory.cpp
```diff
@@ -36,7 +36,7 @@ namespace epiabm
 
     void PopulationFactory::addCell(PopulationPtr population)
     {
-        population->cells().push_back(Cell());
+        population->cells().emplace_back(population->cells().size());
```
KCGallagher, line 39: Not really a comment but more a learning thing for me - why is this better than `push_back()`?

cEpiabm/src/python_bindings/bind_dataclasses.hpp
```diff
@@ -0,0 +1,85 @@
...
+        .def_readwrite("susceptibility", &PersonParams::susceptibility)
+        .def_readwrite("infectiousness", &PersonParams::infectiousness)
+        .def_readonly("next_status_time", &PersonParams::next_status_time);
+    py::class_<Person, PersonPtr>(m, "Person")
+        .def("status", &Person::status, py::return_value_policy::copy)
+        .def("params", &Person::params, py::return_value_policy::reference)
+        .def("cell_pos", &Person::cellPos, py::return_value_policy::copy)
+        .def("microcell_pos", &Person::microcellPos, py::return_value_policy::copy)
+        .def("microcell", &Person::microcell, py::return_value_policy::copy)
+        .def("places", &Person::places, py::return_value_policy::reference)
```
KCGallagher, line 41: Again a learning thing, but what is the distinction between the different `return_value_policy` values here?

cEpiabm/src/simulations/basic_simulation.cpp
```diff
@@ -0,0 +1,89 @@
...
+                    reporter->report(m_population, timestep);
+            }
+        //}
+        // LCOV_EXCL_START
+        /*
+        catch (std::exception& e)
+        {
+            LOG << LOG_LEVEL_ERROR << "Error iterating through timesteps: " << e.what();
+        }*/
+        // LCOV_EXCL_END
```
KCGallagher, line 56: Was this for testing or does it have some function?

cEpiabm/src/sweeps/spatial_sweep.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace epiabm
+{
+
+    SpatialSweep::SpatialSweep() {}
+
+    void SpatialSweep::operator()(const unsigned short timestep)
+    {
+        LOG << LOG_LEVEL_DEBUG << "Beginning Spatial Sweep " << timestep;
+        if (m_population->cells().size() <= 1){
+            return;  // no intercell infections if only one cell
```
KCGallagher, line 21: I wonder if clauses like this should be logged - I think that could be quite useful!

KCGallagher (commented):
Remember us saying the other day we wanted to get this merged in, so thought I would rereview to cover all the new parts I didn't see last time! All looks good, have added a couple of comments to clarify things (and also to help me understand the code as well!)

cEpiabm/src/sweeps/household_sweep.cpp
NicholasFan235: You're right, 0.5 factor just appears in a later commit, probably from testing

cEpiabm/src/dataclasses/person.hpp
NicholasFan235: Commented because they're unused and are technically source implementations in the header file. Not sure what the workaround is, but these lines were causing whole repository to be recompiled even if unrelated files were changed.

cEpiabm/examples/basic_example/main.cpp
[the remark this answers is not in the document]
NicholasFan235: Okay, I see what you mean. It's not a keyword argument, just a comment.
Something like this is done in some of the classes where there is currently an unused parameter and the compiler doesn't like that. So the future parameter is just commented out.

cEpiabm/src/python_bindings/bind_dataclasses.hpp
NicholasFan235: Normally return_value_policy is about ownership i.e. whether python or c++ will delete the object when there are no references to it.
Copy creates a new object in Python with the same value as in C++, and Python has full ownership.
Reference means Python can look at and modify the C++ variable, but when Python is finished with it it won't try to delete it (since it's still a C++ owned object)

There is also return_value_policy::take_ownership
Here the C++ object is given to Python, but Python takes responsibility for deleting it when it's finished with it.

cEpiabm/src/simulations/basic_simulation.cpp
NicholasFan235: Undecided. Is better to catch the error so it can be logged. But for testing with a debugger I needed the error to reach the top level and cause a crash.
TLDR: Was for testing

I-Bouros (approved):
I think it is okay, probably more documentation/comments would be nicer in places though.

cEpiabm/src/population_factory.cpp
NicholasFan235: push_back would construct the Cell (as an rvalue), then copy it into the vector.
emplace_back constructs the Cell in place.

cEpiabm/src/simulations/basic_simulation.cpp
KCGallagher: Does that mean we don't need it anymore, or is it still used in the automated tests? Just aware that's quite a public facing file so commented out code like that might be confusing!

cEpiabm/src/household_linker.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+    class HouseholdLinker
+    {
+    private:
+    public:
+
+        /**
+         * @brief Link households
+         * Create n_households per microcell.
+         * Person has percInHousehold to be in a household.
```
rcw5890, line 20:
```suggestion
         * Person has probability percInHousehold (expressed as percentage) to be in a household.
```

cEpiabm/src/simulations/basic_simulation.hpp
```diff
@@ -0,0 +1,45 @@
...
+        void setup();
+        void teardown();
+    };
+
+    typedef std::shared_ptr<BasicSimulation> BasicSimulationPtr;
+
+}
+
+
+#endif // EPIABM_SIMULATINS_BASIC_SIMULATION_HPP
```
rcw5890, line 45:
```suggestion
#endif // EPIABM_SIMULATIONS_BASIC_SIMULATION_HPP
```

rcw5890 (changes requested):

rcw5890 (approved):

KCGallagher (approved):

KCGallagher:
@NicholasFan235 Is this ready to be merged then?

PR #41, opened 2022-01-14
Py restructure

### LINKED ISSUE #25: Restructure pyEpiabm repo
Best left until the initial simulation milestone is reached, so that no one has active code on their remotes to avoid many horrific conflicts

KCGallagher:
The first two commits only have files being moved so can be ignored to help see the actual changed for reviewing! 

I have tried to tidy up import lists etc as well (mirroring the styles used within [scipy](https://github.com/scipy/scipy/blob/main/scipy/stats/_kde.py) and [PEP8](https://www.python.org/dev/peps/pep-0008/#imports)).

Closes #25


pyEpiabm/pyEpiabm/__init__.py
```diff
@@ -1,44 +1,18 @@
...
-
-# Example Simulations
-from .simulation import Simulation
+# from pyEpiabm.property.infection_status import InfectionStatus
 
-# Data collection
-from ._compartment_counter import _CompartmentCounter
-from ._csv_dict_writer import _CsvDictWriter
-from ._csv_writer import _CsvWriter
+from . import core
```
rcw5890, line 14:
```suggestion
from . import core
from .core.cell import Cell
```
etc.
Did you consider making the core classes available right under pyepiabm for ease of access?
It's fine as it is, but I would also be happy if core was just a category for our internal use and its classes were available directly.

rcw5890 (changes requested):
Thanks for making these improvements

pyEpiabm/pyEpiabm/__init__.py
```diff
@@ -1,44 +1,18 @@
...
-from .host_progression_sweep import HostProgressionSweep
-from .household_sweep import HouseholdSweep
-from .initial_infected_sweep import InitialInfectedSweep
-from .place_sweep import PlaceSweep
-from .queue_sweep import QueueSweep
-from .update_place_sweep import UpdatePlaceSweep
-
-# Example Simulations
-from .simulation import Simulation
+# from pyEpiabm.property.infection_status import InfectionStatus
```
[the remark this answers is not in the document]
KCGallagher: Don't know how I missed this - thanks!

pyEpiabm/pyEpiabm/__init__.py
KCGallagher: Hmm, how far would we extend this? I thought one of the main motivations in reorganising the init files specifically was to avoid cluttering the pyEpiabm namespace (so we wouldn't want everything in there) and I think it could become hard to decide what is important enough to be in the pyEpiabm namespace (I know this is in fact an issue larger modules like scipy come across).

Personally I think I tend towards leaving it as is, rather than trying to make an arbitrary decision on which modules go in the top level init and introducing inconsistency on import methods, however very happy to chat about this @rcw5890 ?

rcw5890: > Hmm, how far would we extend this? I thought one of the main motivations in reorganising the init files specifically was to avoid cluttering the pyEpiabm namespace (so we wouldn't want everything in there) and I think it could become hard to decide what is important enough to be in the pyEpiabm namespace (I know this is in fact an issue larger modules like scipy come across).
> 
> Personally I think I tend towards leaving it as is, rather than trying to make an arbitrary decision on which modules go in the top level init and introducing inconsistency on import methods, however very happy to chat about this @rcw5890 ?

It's not at all unprecedented to have a `core` subpackage but expose all its contents in the main namespace---as far as I know this is exactly what numpy, pandas, and sympy are all doing. I agree we will be repeatedly asking whether something is central enough to go there, but that's not insurmountable.
scipy is more an exception owing to its extremely broad subject coverage. They don't seem to put anything at all in the main namespace except scipy.test(), but all their subpackages are meaningful subcategories covering a specific topic much narrower in scope than scipy itself.

Though I don't disagree that not touching the main namespace has the appeal of consistency, so I'm happy either way---PR approved and I'd gladly hear any more ideas on this topic.

rcw5890 (approved):

pyEpiabm/pyEpiabm/__init__.py
KCGallagher: Changed as requested - you can now import Cell from pyEpiabm directly (although not when within the pyEpiabm code itself as that would result in circular imports)

PR #43, opened 2022-01-19
Py infection force

### LINKED ISSUE #37: pyEpiabm - sort out CovidSim helpers class
Class where we originslly shoved all the calculate inf and susc for places/households etc. Depending on how heavily we want to copy Covidsim may need refactoring

KCGallagher:
Tidy up covidhelpers class by creating separate files for the place and household methods. Decided to keep this in routines, as it is not a property of a particular class, but rather depends on a number of instances of different classes in the current population. It could depend on population wide attributes in the future as well.

Closes #37.


pyEpiabm/pyEpiabm/routine/household_forces.py
```diff
@@ -0,0 +1,57 @@
...
+        :param timestep: Current simulation timestep
+        :type timestep: int
+        :return: Susceptibility parameter of household
+        :rtype: float
+        """
+        return 0.1
+
+    @staticmethod
+    def household_inf_force(infector: Person, infectee: Person, timestep: int):
+        """Calculate the force of infector parameter of a household,
```
KCGallagher, line 42:
```suggestion
        """Calculate the force of infection parameter of a household,
```

KCGallagher (commented):
One typo to fix

pyEpiabm/pyEpiabm/routine/household_forces.py
```diff
@@ -0,0 +1,57 @@
...
+        :type infectee: Person
+        :param timestep: Current simulation timestep
+        :type timestep: int
+        :return: Susceptibility parameter of household
+        :rtype: float
+        """
+        return 0.1
+
+    @staticmethod
+    def household_inf_force(infector: Person, infectee: Person, timestep: int):
```
Elizabeth-Hayman, line 41: Seeing as this calculates the infectiousness*susceptibility maybe household_FOI would be a better name to avoid confusion with the household_inf one

pyEpiabm/docs/source/routines.rst
```diff
@@ -4,15 +4,19 @@ Routines
 
 .. currentmodule:: pyEpiabm.routine
 
-Routines provides various methods to act upon or create a population.
+Routines provides various methods to act upon or create a population, or members of it.
 
 Overview:
 
-- :class:`CovidsimHelpers`
+- :class:`HouseholdForces`
```
Elizabeth-Hayman, line 11: I feel HouseholdForces is a bit of a weird name. It suggests physical forces. Could we make FOI for force of infection well used and described throughout the code (a la covidsim) so HouseholdFOI and calc_household_foi etc, see below

pyEpiabm/pyEpiabm/routine/place_forces.py
```diff
@@ -0,0 +1,65 @@
...
+        :type place: Place
+        :param timestep: Current simulation timestep
+        :type timestep: int
+        :return: Infectiousness parameter of place
+        :rtype: float
+        """
+        return 0.5
+
+    @staticmethod
+    def place_inf_force(place: Place, infector: Person, infectee: Person,
```
Elizabeth-Hayman, line 46: Thought for later: we probably want to have a baseline susc, inf, foi and then possibility of changing it within a simulation (ie if travel restrictions come in foi =0 etc). Could we have a  place_base_foi and separate method place_adapt_foi (and maybe another separate place_reset_foi) which would take the baseline and change it

Elizabeth-Hayman (commented):
I think this refactoring makes sense, and will make it easier to adjust infection rates later.

pyEpiabm/pyEpiabm/routine/household_forces.py
Elizabeth-Hayman: actually household_foi, sorry

KCGallagher: Yeah I like this, wasn't massively happy with the original name (even if I read _'foi'_ as freedom of information)

pyEpiabm/docs/source/routines.rst
KCGallagher: @Elizabeth-Hayman  Hmm this is the name for the class though, which includes other things too? Or were you saying that all the other factors that Ferguson later adds just modify the force of infection? I am certainly amenable to changing this!

pyEpiabm/pyEpiabm/routine/place_forces.py
KCGallagher: I agree, but think this might be too much future engineering; these won't be hard to add later so better to do it then when we know what we want!

pyEpiabm/docs/source/routines.rst
KCGallagher: I decided to go for `HouseholdInfection` etc as the class (wanted to avoid using an acronym in the class name and thought we might want to extend the class later) - hope this is okay!

Elizabeth-Hayman: Yep this is good

Elizabeth-Hayman (approved):
Tis all good

pyEpiabm/pyEpiabm/routine/household_infection.py
```diff
@@ -0,0 +1,57 @@
...
+from pyEpiabm.core import Person
+
+
+class HouseholdInfection:
+    """Class to calculate the infectiousness and susceptibility
+    parameters for the force of infection parameter, within households.
+    """
+    @staticmethod
+    def household_inf(infector: Person, timestep: int):
+        """Calculate the infectiveness of a household.
```
rcw5890, line 14:
```suggestion
        """Calculate the infectiousness of a household.
```
Saves us from having to make extra changes on the other PR

pyEpiabm/pyEpiabm/routine/household_infection.py
```diff
@@ -0,0 +1,57 @@
...
+
+        :param infector: Infector
+        :type infector: Person
+        :param timestep: Current simulation timestep
+        :type timestep: int
+        :return: Infectiveness parameter of household
```
rcw5890, line 20:
```suggestion
        :return: Infectiousness parameter of household
```

rcw5890 (changes requested):

rcw5890 (approved):

PR #44, opened 2022-01-24
Py random seed

### LINKED ISSUE #38: pyEpiabm - set random seed
Ensure the model can be ran in a fully deterministic manner for research purposes by having an option to set the random seed at the start of the simulation. If only python and numpy random modules are used throughout, this should be fairly easy to do.

@NicholasFan235  you might want to consider implementing something similar in the C++ version?

KCGallagher:
# PR - Set the random seed of pyEpiabm simulation 
I have added the option to set the random seed for reproducibility in the simulation. Closes #38, #45

## Notes
There are actually two separate seeds, one for generating the population and one for running the simulation. I am aware this is a little more work, but I think this is desirable as we may want to run multiple random simulations on the same initial population. Both are included in the parameters dictionary within the simultion workflow.

## Testing
This was tricky. I have come up with a complicated test for the population seed to ensure that setting the same seed results in the same (random) allocation of individuals across cells/microcells, but using a different seed results in a different allocation.

I am struggling to think of a way to unit test the simulation as a whole (without dependency on each of the individual parts) but don't want to write seed tests for any function with includes a random element. It might be best to run two simulations with the same seed and verify they are identical, although even this is non-trivial to write (do you attempt to compare all numerical outputs of the simulation?) Suggestions/contributions welcome on this matter!


KCGallagher:
Update: Have written a proper test for testing the simulation seed - its not a unit test though because it runs the entire simulation. It compares the final state of the simulation, and checks it is the same when the seed is the same, but different when the seed is changed.

Naturally, this requires the final state to be unique, and assumes the simulation does not run to completion. I have set the parameters to ensure this is the case currently, but I am aware this might change as other classes in the simulation are updated. If anyone has a better way of testing this simulation seed however, that would be most appreciated!

rcw5890 (changes requested):
@KCGallagher thanks very much for adding this.
My main concern with the current implementation is that it sets the seed during configure/make_pop, but it never does anything to maintain it in between then and the methods that use random numbers being called. So if someone did
```
sim = Simulation(seed1)
pop2 = ToyPopulationFactory.make_pop(seed2)
sim.run_sweeps()
```
sim would no longer be obeying seed1 and reproducibility would be lost.

I haven't given it much thought beyond this yet and would be happy to hear any ideas. One idea:---
```python
class:
    def __init__(seed):
        self.py_random = random.Random(seed)
        self.np_random = np.random.default_rng(seed)

    def method():
        sample_from_python = self.py_random.random()
        sample_from_numpy = self.np_random.random()
```
The downside of this approach is that self.py_random and self.np_random then have to be passed around to other objects---in our case Simulation would have to send them to the sweeps. This could be done by making them optional arguments to the sweep call method, or adding a bind_simulation similar to bind_population so that the sweeps could access their simulation.

KCGallagher:
> @KCGallagher thanks very much for adding this. My main concern with the current implementation is that it sets the seed during configure/make_pop, but it never does anything to maintain it in between then and the methods that use random numbers being called. So if someone did
> 
> ```
> sim = Simulation(seed1)
> pop2 = ToyPopulationFactory.make_pop(seed2)
> sim.run_sweeps()
> ```
> 
> sim would no longer be obeying seed1 and reproducibility would be lost.
> 
> I haven't given it much thought beyond this yet and would be happy to hear any ideas. One idea:---
> 
> ```python
> class:
>     def __init__(seed):
>         self.py_random = random.Random(seed)
>         self.np_random = np.random.default_rng(seed)
> 
>     def method():
>         sample_from_python = self.py_random.random()
>         sample_from_numpy = self.np_random.random()
> ```
> 
> The downside of this approach is that self.py_random and self.np_random then have to be passed around to other objects---in our case Simulation would have to send them to the sweeps. This could be done by making them optional arguments to the sweep call method, or adding a bind_simulation similar to bind_population so that the sweeps could access their simulation.

@rcw5890  Thanks for your comments! I have to admit I am rather against passing them around as arguments - I think this could get rather unwieldy and wanted it to be a process occurring in the background! The other issue with the class is that we often want different methods from numpy (ie a random integers, floats or arrays) and so the class above could become more complex.

Maybe this is a disadvantage of having two separate seeds - after all if someone wants to run multiple simulations on one population they can do so by configuring one population in advance and then iterating over copies of it, or just being careful in the order of operations? Or indeed there is no need to have to set different seeds, just that you can either set a fixed seed for the population only, or the same value for both the population and the simulation? 

In any case, clearly a complicated issue and one worth discussing tomorrow!

rcw5890 (approved):

PR #47, opened 2022-01-26
spatial sweep

Elizabeth-Hayman:
Spatial sweep created in the style of Covidsim's version to look at inter-cell infections. 

For each cell Covidsim uses a Poisson variable to decide how many infection events may occur.  Then it decides on a cell the infectee belongs in, then specifically on the infector, infectee.

Apart from tests, the one extra addition is a function in the cell class to calculate the total number of infectious people in each cell


pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -0,0 +1,93 @@
...
+                # force of infection specific to cells and people
+                # involved in the infection event
+                force_of_infection = SpatialInfection.\
+                    space_foi(cell, infectee_cell, infector, infectee)
+
+                # Compare a uniform random number to the force of
+                # infection to see whether an infection event
+                # occurs in this timestep between the given
+                # persons.
+                r = random.uniform(0, 1)
```
KCGallagher, line 90: Any reason for this and not just the default random.random()?

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+
+import pyEpiabm as pe
+from pyEpiabm.property.infection_status import InfectionStatus
```
KCGallagher, line 4:
```suggestion
from pyEpiabm.property import InfectionStatus
```

May need to add to __init__.py in thei subpackage if this doesn't work yet, but should be consistent with the other import!

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
```diff
@@ -0,0 +1,52 @@
...
+        """Intialise a population with one infector and one
+        infectee, both in the same place and household.
+        """
+        cls.cell = pe.Cell()
+        cls.microcell = pe.Microcell(cls.cell)
+        cls.infector = pe.Person(cls.microcell)
+        cls.infectee = pe.Person(cls.microcell)
+        cls.timestep = 1
+
+    def test_calc_place_susc(self):
```
KCGallagher, line 25:
```suggestion
    def test_calc_space_susc(self):
```

I'm not saying this was a copy and paste job but you should maybe change the test names! If you wouldn't mind doing me a favour and removing '_calc' from each test name in this file and `test_place_forces.py` that would be much appreciated!

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -0,0 +1,69 @@
...
+        self.cell_susc.add_microcells(1)
+        self.microcell_susc = self.cell_susc.microcells[0]
+        self.microcell_susc.add_people(1)
+        self.infectee = self.microcell_susc.persons[0]
+
+        test_sweep.bind_population(self.pop)
+        test_sweep(time)
+        self.assertTrue(self.cell_susc.person_queue.empty())
+
+        # Change person"s status to infected
```
KCGallagher, line 60:
```suggestion
        # Change person's status to infected
```

KCGallagher (changes requested):
This is so good, can't believe you did it all in an afternoon! Mostly superficial changes here (I can't really judge how well you have implemented CovidSim as I don't know that section of the code) but it all looks sensible

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
Elizabeth-Hayman: oops so att-ed

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
Elizabeth-Hayman: nope, changed :)

pyEpiabm/pyEpiabm/routine/spatial_infection.py
```diff
@@ -0,0 +1,88 @@
...
+        :rtype: float
+        """
+        return 0.2
+
+    @staticmethod
+    def space_inf(inf_cell: Cell, infector: Person,
+                  timestep: int):
+        """Calculate the infectiousness between cells.
+        Dependent on the infectious people in it, and social
+        distancing measures.
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Absolutely sure

pyEpiabm/pyEpiabm/routine/spatial_infection.py
```diff
@@ -0,0 +1,88 @@
...
+        towards its neighbouring cells.
+
+        :param inf_cell: cell doing infecting
+        :type inf_cell: Cell
+        :param timestep: Current simulation timestep
+        :type timestep: int
+        :return: Average number of infection events from the cell
+        :rtype: int
+        """
+        R_value = 2  # R_value might be the wrong term
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Yea completely. This should be the amount of people each individual infected is going to infect WITHOUT any restrictions on their behaviour (not sure thats quite the strict definition of R value?).  So only dependent on the virus. Then we can add in NPIs to this secticon later (not sure thats quite the strict definition of R value?).

Elizabeth-Hayman:
All the failures are being caused by an import error, which suggests I haven't updated an init somewhere. The only import I've changed during this PR was property.InfectionStatus to property (and also haven't updated the docs but that didn't seem to fail?)

pyEpiabm/pyEpiabm/routine/spatial_infection.py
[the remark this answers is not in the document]
KCGallagher: So does that make it a simulation parameter passed in at the start of the simulation?

KCGallagher:
> All the failures are being caused by an import error, which suggests I haven't updated an init somewhere. The only import I've changed during this PR was property.InfectionStatus to property (and also haven't updated the docs but that didn't seem to fail?)

@Elizabeth-Hayman Hmm this is an interesting one. I can see the issue - the routine subpackage requires simulation to configure, which requires various sweeps. But then trying to configure the sweep subpackage, it runs into issues with your spatial sweep, which itself relies on the routine subpackage, so neither subpackage can be configured without the other. What surprises me more is that it is still fine with the household sweep, which also relies on the routine subpackage. 

Given it was working before my suggestions, I feel some responsibility here I don't think it can be the import change as that was in the tests so shouldn't affect the module itself. While the tests seem to be failing from the first commit - I suspect it is the import reordering instead that I suggested!

KCGallagher:
Seems like the issue was pretty nuanced, and I don't like that it happened. Reordering the imports means that the rest of the routine sub-package can be configured (including the spatial sweep) before the simulation reference is called, preventing any circular import.

Of course, we will have to reorder imports by inheritance hierarchy in future, but I am worried that will be hard to keep track of - @rcw5890 do you think there is a better solution or is that inevitable in a complex (and interdependent) package like this, without having to configure all classes in one import level? I am aware we have based our import structure on analysis packages like scipy which have a much more linear dependence hierarchy than a simulation package.

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
```diff
@@ -22,19 +22,19 @@ def setUpClass(cls) -> None:
         cls.infectee = pe.Person(cls.microcell)
         cls.timestep = 1
 
-    def test_calc_place_susc(self):
+    def test_place_susc(self):
```
KCGallagher, line 25:
```suggestion
    def test_space_susc(self):
```

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
```diff
@@ -22,19 +22,19 @@ def setUpClass(cls) -> None:
...
         result = SpatialInfection.space_susc(self.cell, self.infectee,
                                              self.timestep)
         self.assertTrue(result > 0)
         self.assertIsInstance(result, float)
 
-    def test_calc_place_inf(self):
+    def test_place_inf(self):
```
KCGallagher, line 31:
```suggestion
    def test_space_inf(self):
```

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
```diff
@@ -22,19 +22,19 @@ def setUpClass(cls) -> None:
...
         result = SpatialInfection.space_inf(self.cell, self.infector,
                                             self.timestep)
         self.assertTrue(result > 0)
         self.assertIsInstance(result, float)
 
-    def test_calc_place_foi(self):
+    def test_place_foi(self):
```
KCGallagher, line 37:
```suggestion
    def test_space_foi(self):
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertEqual(self.cell_susc.person_queue.qsize(), 1)
+
+        # Add multiple people to infector cell
+        self.cell_inf.microcells[0].add_people(10000)
+        self.cell_susc.person_queue = Queue()
+        test_sweep.bind_population(self.pop)
+        test_sweep(time)
+        self.assertEqual(self.cell_susc.person_queue.qsize(), 1)
+
+        # Change infectee"s status to recovered
```
rcw5890, line 74:
```suggestion
        # Change infectee's status to recovered
```

rcw5890 (changes requested):
Thanks @Elizabeth-Hayman  it looks great

rcw5890:
> Seems like the issue was pretty nuanced, and I don't like that it happened. Reordering the imports means that the rest of the routine sub-package can be configured (including the spatial sweep) before the simulation reference is called, preventing any circular import.
> 
> Of course, we will have to reorder imports by inheritance hierarchy in future, but I am worried that will be hard to keep track of - @rcw5890 do you think there is a better solution or is that inevitable in a complex (and interdependent) package like this, without having to configure all classes in one import level? I am aware we have based our import structure on analysis packages like scipy which have a much more linear dependence hierarchy than a simulation package.

As much as possible let's avoid having to order the imports in a particular way for the code to run!

We can get rid of the ImportError (even without your change to `routine/__init__.py`) by in spatial_sweep.py replacing `from pyEpiabm.routine import SpatialInfection` with `import pyEpiabm.routine`. Then, you have to access SpatialInfection using pyEpiabm.routine.SpatialInfection.

Apparently the `from ... import something` syntax is more sensitive than just importing the subpackage but not trying to access anything within it until the code is run. We probably want to change this everywhere we have a circular import.

KCGallagher:
> > Seems like the issue was pretty nuanced, and I don't like that it happened. Reordering the imports means that the rest of the routine sub-package can be configured (including the spatial sweep) before the simulation reference is called, preventing any circular import.
> > Of course, we will have to reorder imports by inheritance hierarchy in future, but I am worried that will be hard to keep track of - @rcw5890 do you think there is a better solution or is that inevitable in a complex (and interdependent) package like this, without having to configure all classes in one import level? I am aware we have based our import structure on analysis packages like scipy which have a much more linear dependence hierarchy than a simulation package.
> 
> As much as possible let's avoid having to order the imports in a particular way for the code to run!
> 
> We can get rid of the ImportError (even without your change to `routine/__init__.py`) by in spatial_sweep.py replacing `from pyEpiabm.routine import SpatialInfection` with `import pyEpiabm.routine`. Then, you have to access SpatialInfection using pyEpiabm.routine.SpatialInfection.
> 
> Apparently the `from ... import something` syntax is more sensitive than just importing the subpackage but not trying to access anything within it until the code is run. We probably want to change this everywhere we have a circular import.

Ahh yes I had heard of this before and should have remembered - that's definitely a better solution!

pyEpiabm/pyEpiabm/routine/spatial_infection.py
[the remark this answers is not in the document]
Elizabeth-Hayman: Have changed to a parameter class instance

I-Bouros (approved):

pyEpiabm/pyEpiabm/routine/spatial_infection.py
```diff
@@ -0,0 +1,88 @@
...
+        :type timestep: int
+        :return: Average number of infection events from the cell
+        :rtype: int
+        """
+        R_0 = Parameters.instance().basic_reproduction_num
+        total_infectors = inf_cell.number_infectious()
+        # Add in other classes of people who are infectors
+        number_to_infect = total_infectors * R_0
+        # Round this float to nearest integer
+        return round(number_to_infect)
```
KCGallagher, line 29: I'm not certain this is the best behaviour - say this is 1.5 then all cells would have value 2. Surely we would want half of cells to have value 1 and half to have value 2 (randomly assigned)? Do we know how CovidSim does it?

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -0,0 +1,93 @@
...
+        :type time: int
+        """
+        timestep = int(time * Parameters.instance().time_steps_per_day)
+        # Double loop over the whole population, checking infectiousness
+        # status, and whether they are absent from their household.
+        for cell in self._population.cells:
+
+            # As this tracks intercell infections need to check number of
+            # cells is more than one (edge case but worth having)
+            if len(self._population.cells) == 1:
```
KCGallagher, line 43: Is there a way to have this outside the for loop, and just break the function call at that point? Would just save checking this for each cell, but it's not a big extra cost!

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertTrue(result > 0)
+        self.assertIsInstance(result, float)
+
+    def test_space_inf(self):
+        result = SpatialInfection.space_inf(self.cell, self.infector,
+                                            self.timestep)
+        self.assertTrue(result > 0)
+        self.assertIsInstance(result, float)
+
+    def test_space_foi(self):
```
KCGallagher, line 38: Do we require this to be less than one as well, or can it be greater than 1 too (resulting in guaranteed infection)?

KCGallagher (commented):
Have left some comments on sections I'm not sure about - may not need changes but just wanted to raise a few things to make sure we'd thought about them!

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
Elizabeth-Hayman: I like allowing it to be any non-negative number, as I feel an infectiousness of 3 is very different to one of 1, even though they both lead to the same number of people infected.

pyEpiabm/pyEpiabm/routine/spatial_infection.py
Elizabeth-Hayman: Implemented with choice weighted by how close the float is to respective neighbouring integers

pyEpiabm/pyEpiabm/tests/test_routine/test_space_forces.py
Elizabeth-Hayman: I like allowing it to be any non-negative number, as I feel an infectiousness of 3 is very different to one of 1, even though they both lead to the same number of people infected.

KCGallagher: I am now convinced that the force of infection may be greater than 1 and so it may be beneficial to allow this in the future!

KCGallagher (approved):

rcw5890 (approved):

PR #52, opened 2022-02-02
Py mock issue

### LINKED ISSUE #51: Fix mocks in testing
Running the tests generates an empty directory in the top level - I think this is from my spatial logger but currently isolating the problem to ensure this folder generation is mocked properly to avoid this

KCGallagher:
I have mocked the `mkdir` call to prevent spurious folder generation in tests.

I have also removed the handling of the file not found error, as I can find a way to trigger this! Previously this handled issues where the folder didn't exist, but this is no longer an issue as the abstract reporter will create one. There will not be issues with the file not existing, as it is opened in write mode, so one will be created if it doesn't exist. If you can think of a case which isn't handled here let me know, but otherwise I think we no longer need it.

Closes #51.

### DIFF
--- a/pyEpiabm/pyEpiabm/output/_csv_dict_writer.py
+++ b/pyEpiabm/pyEpiabm/output/_csv_dict_writer.py
@@ -25,16 +25,10 @@ class _CsvDictWriter(AbstractReporter):
         """
         super().__init__(folder, clear_folder)
 
-        try:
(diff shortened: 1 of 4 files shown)


Elizabeth-Hayman:
Looks beautiful

rcw5890 (approved):

PR #58, opened 2022-02-15
Spatial weighting

Elizabeth-Hayman:
Added in a distance metric and popped it into the spatial sweep. (Currently implemented as cells being chosen based on a weighting which is the inverse of the distance between them).


pyEpiabm/pyEpiabm/tests/test_core/test_cell.py
```diff
@@ -39,6 +48,12 @@ def test_number_infectious(self):
         person.update_status(InfectionStatus.InfectMild)
         self.assertEqual(cell.number_infectious(), 1)
 
+    def test_set_loc(self):
+        cell = pe.Cell()
+        self.assertEqual(cell.location, (0, 0))
+        cell.set_location((3.0, 3.0))
+        self.assertEqual(cell.location, (3.0, 3.0))
```
KCGallagher, line 55:
```suggestion
        cell.set_location((3.0, 2.0))
        self.assertEqual(cell.location, (3.0, 2.0))
```

This might be my code, but correcting my own mistakes it would definitely be safer to have two different values in this test (just in case the indices have been mixed up in the function that sets the location)

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -2,19 +2,21 @@
 from unittest import mock
 from queue import Queue
 
-import pyEpiabm as pe
+from pyEpiabm.core import Population, Parameters
+from pyEpiabm.property import InfectionStatus
+from pyEpiabm.sweep import SpatialSweep
```
KCGallagher, line 7: Not a major thing, but it is very fine to import the whole of pyEpiabm in the tests (and probably a good thing to do at least some of the time to check everything can be accessed in the expected way :)

pyEpiabm/pyEpiabm/tests/test_utility/test_distance_metrics.py
```diff
@@ -0,0 +1,22 @@
...
+    def test_dist_euclid(self):
+        dist = DistanceFunctions.dist_euclid((3, 0))
+        self.assertTrue(dist == 3)
+
+        dist = DistanceFunctions.dist_euclid((-3, 0))
+        self.assertTrue(dist == 3)
+
+        dist = DistanceFunctions.dist_euclid((-3, 0), (3, 0))
+        self.assertTrue(dist == 6)
+
```
KCGallagher, line 19:
```suggestion
        dist = DistanceFunctions.dist_euclid((-3, 0), (3, 0))
        self.assertTrue(dist == 6)

        dist = DistanceFunctions.dist_euclid((-2, 0), (1, 4))
        self.assertTrue(dist == 5.0)
```
I think it would be good to have at least one check for the y coordinates as well - this is just a suggestion and I haven't actually ran it to verify it works! Also I think the function should probably return a float?

python_examples/simulation_flow.py
```diff
@@ -10,26 +10,26 @@
 
 # Method to set the seed at the start of the simulation, for reproducibility
 
-pe.routine.Simulation.set_random_seed(seed=42)
+# pe.routine.Simulation.set_random_seed(seed=42)
```
KCGallagher, line 13: Any reason for this change? Was useful to control output graph for readme (and similarly useful to keep the same output)

KCGallagher (changes requested):
Overall looks really good - a few minor changes in comments.

One larger suggestion - I think it would be good to discard your changes from the example simulation workflow, as this is meant to be the simplest possible example for users to look at without having to consider the more complex spatial elements of this model.

Instead if you could rerun the newer spatial plotters (should be merged into main imminently) with this spatial infection, that should produce nicer (and much more informative plots). This would also mean discarding the changes to the simulation flow output .csv and plots, and instead submitting new plots within the spatial example directory (appropriately explained in the readme!)

Sorry this is a pain, but it should help give a much better overall picture of what this new feature is doing, and visually represent its impact in the voronoi plots so we can see what difference it makes!

I-Bouros (changes requested):

pyEpiabm/python_examples/simulation_outputs/output.csv
```diff
@@ -0,0 +1,61 @@
+time,InfectionStatus.Susceptible,InfectionStatus.Exposed,InfectionStatus.InfectASympt,InfectionStatus.InfectMild,InfectionStatus.InfectGP,InfectionStatus.InfectHosp,InfectionStatus.InfectICU,InfectionStatus.InfectICURecov,InfectionStatus.Recovered,InfectionStatus.Dead
```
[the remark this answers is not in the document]
KCGallagher: Ahh think I submitted this as you delete the .csv file!

KCGallagher:
I have removed misc output files, will have a play to generate some new pretty graphs this week before re-requesting review.

KCGallagher (commented):
This looks really good, and love the `if Parameters.instance().do_CovidSim:` line! Have we put these two different versions into the Wiki?
You might hate me for this suggestion but do you fancy translating these latest changes (Wed 16th pm) into C++ after this PR is completed - it should be quite straight forward as I have done a lot of the groundwork for spatial sweeps and would be a nice intro to the C++ code I think!

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -42,52 +44,96 @@ def __call__(self, time: float):
...
+                                                   weights=weights, k=1)[0]
+                    # Sample at random from the infectee cell to find
+                    # an infectee
+                    infectee = random.sample(infectee_cell.persons, 1)[0]
+                    infection_distance = DistanceFunctions.dist_euclid(
+                     cell.location, infectee_cell.location) / Parameters.\
+                        instance().infection_radius
+                    if (infection_distance < random.random()):
+                        # Covidsim rejects the infection event if the distance
+                        # between infector/infectee is too large.
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Line above? Divides through by another Parameter instance which I'm taking to be a distance

Elizabeth-Hayman:
To prevent hving to call setup all the time in tests (which wasn't great anyway for testing as it assumes everyone is susceptible so you had to sort them out manually) I added a super short add_new_person to compartment counter, and that is called every time people are added to cells/microcells. Think this renders all my afternoon changes to the core test redundant but will think about that in the morning.

pyEpiabm/pyEpiabm/core/_compartment_counter.py
```diff
@@ -47,9 +47,17 @@ def report(self, old_status: InfectionStatus,
         :param new_status: Person's new infection state
         :type new_status: InfectionStatus
```
KCGallagher, line 48:
```suggestion
        :param new_status: Person's new infection state
        :type new_status: InfectionStatus
        :param new_person: Whether person has just been added to population
        :type new_person: bool
```

pyEpiabm/pyEpiabm/core/_compartment_counter.py
```diff
@@ -38,7 +38,7 @@ def initialize(self, n_people) -> None:
         self._compartments[InfectionStatus.Susceptible] = n_people
 
     def report(self, old_status: InfectionStatus,
-               new_status: InfectionStatus) -> None:
+               new_status: InfectionStatus, new_person=False) -> None:
```
KCGallagher, line 41: Is this variable used anywhere?

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -42,52 +44,97 @@ def __call__(self, time: float):
...
+
+                distance_weights = [1/DistanceFunctions.dist_euclid(
+                                    cell.location, cell2.location)
+                                    for cell2 in
+                                    possible_infectee_cells]
+                cell_list = random.choices(possible_infectee_cells,
+                                           weights=distance_weights,
+                                           k=number_to_infect)
+                print(cell_list)
+
```
KCGallagher, line 109:
```suggestion
```

Think this is left from debugging?

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -42,52 +44,97 @@ def __call__(self, time: float):
...
+                # Each infection event corresponds to a infectee cell
+                # on the cell list
+                for infectee_cell in cell_list:
+
+                    # Sample at random from the infectee cell to find
+                    # an infectee
+                    infectee = random.sample(infectee_cell.persons, 1)[0]
+                    self.do_infection_event(infector, infectee, timestep)
+
+    def do_infection_event(self, infector, infectee,
```
KCGallagher, line 119: If this method isn't private (and maybe even if it is) it should be unit tested - not that I'm trying to get my tests re-added ;)

KCGallagher (changes requested):
I can see we want to remove the use of `_setup()` from the tests, but surely `pop.setup()` is the correct usage we want to use/encourage?

pyEpiabm/pyEpiabm/core/_compartment_counter.py
Elizabeth-Hayman: OutDATED NOW, SOZ

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
Elizabeth-Hayman: Ooops, thanks

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
Elizabeth-Hayman: Hah, okey, I can separate that out

Elizabeth-Hayman:
Would be very happy to work on this in C++, will also update the wiki soonish

pyEpiabm/python_examples/simulation_outputs/output.csv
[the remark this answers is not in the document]
Elizabeth-Hayman: gosh

KCGallagher:
Used the new sweep to create some spatial plots based on a single initial infected individual. All FOIs haven't been initialised properly yet so it spreads between cells much faster than you might expect. No significant spatial variation on this scale - I was hoping to be able to visualise a wave of virus spreading spatially but you can't set the cut-off distance for transmission this low as it can then get stuck with no-one in nearby cells to infect (I wonder if Covidsim has any way of handling this, or just relies on susceptible individuals in nearby cells, whereas currently everyone in affected cells gets infected in our simulation, unlike real life). 

Might be nice to consider a breakout clause for the do while loop in spatial infection, say when every cell within the cut-off radius has been considered, as it took me a while to figure out what was going on here!

pyEpiabm/pyEpiabm/core/_compartment_counter.py
```diff
@@ -47,9 +53,17 @@ def report(self, old_status: InfectionStatus,
         :param new_status: Person's new infection state
         :type new_status: InfectionStatus
         """
+        if self._compartments[old_status] == 0:
+            raise ValueError("No people of this status in this cell.")
         self._compartments[old_status] -= 1
         self._compartments[new_status] += 1
 
+    def report_new_person(self) -> None:
```
NicholasFan235, line 61: Is this function just here for setting up the population? We're not actually adding new people mid-simulation?
The change you made to initialize is perfect, and in the C++ I just have it called by the simulation class right before starting the iterations. I don't have any problem doing it this way instead, but should maybe name it something else or at least document if it's just for initialization

pyEpiabm/pyEpiabm/utility/distance_metrics.py
```diff
@@ -0,0 +1,30 @@
...
+
+import typing
+import numpy as np
+
+
+class DistanceFunctions:
+    """Class which contains multiple distance functions
+    for use when considering spatial infections, either
+    inter-place or inter-cell.
+    """
```
NicholasFan235, line 13: What if we just create a staticmethod dist here which then uses dist_euclid. Will make it easier to change method in the future.

pyEpiabm/pyEpiabm/core/_compartment_counter.py
```diff
@@ -28,17 +28,23 @@ def identifier(self):
...
-        self._compartments[InfectionStatus.Susceptible] = n_people
+        # Not sure how the docs work if it accepts two input types
+        # First reset to zeros. This may seem unnecessary, but will
+        # prevent duplicate code double-counting.
+        self._compartments = {status: 0 for status in InfectionStatus}
+        for person in cell.persons:
+            inf_status = person.infection_status
+            self._compartments[inf_status] += 1
 
     def report(self, old_status: InfectionStatus,
```
KCGallagher, line 46: Remove handling for new person here (in docs too)

pyEpiabm/pyEpiabm/core/_compartment_counter.py
Elizabeth-Hayman: Think I replied on teams:

I kept add new person separate primarily for the testing where you're changing the population around a lot its needed. Can definitely see use in a full simulation though, maybe if infected people are coming into the country from abroad. Hopefully none of the compartment counters should be called explicitly now, it's all wrapped up in either add_person to microcell of in the population factory. So will document that

pyEpiabm/pyEpiabm/core/_compartment_counter.py
```diff
@@ -47,9 +53,17 @@ def report(self, old_status: InfectionStatus,
         :param new_status: Person's new infection state
         :type new_status: InfectionStatus
         """
+        if self._compartments[old_status] == 0:
```
KCGallagher, line 56:
```suggestion
        if self._compartments[old_status] <= 0:
```

Should never be negative anyway (in which case it will be equivalent to the old version) but just in case?

pyEpiabm/pyEpiabm/routine/toy_population_config.py
```diff
@@ -78,7 +78,7 @@ def make_pop(pop_params: typing.Dict):
         if place_number > 0:
             ToyPopulationFactory.add_places(new_pop, place_number)
 
-        new_pop.setup()
+        # new_pop.setup()
```
KCGallagher, line 81:
```suggestion
```

KCGallagher: If it needs to go it needs to go!

pyEpiabm/pyEpiabm/tests/test_core/test_compartment_counter.py
```diff
@@ -26,10 +39,16 @@ def test_reportRetrieve(self):
...
 
     def test_reportRetrieveLarge(self):
+        # Would note this takes a while to run so slows down unittests
         subject = pe._CompartmentCounter("")
-        subject.initialize(1000000)
+        large_num = 1000000
+        self.microcell.add_people(large_num - len(self.microcell.persons))
+        subject.initialize(self.microcell)
         statuses = {s: 0 for s in pe.property.InfectionStatus}
         statuses[pe.property.InfectionStatus.Susceptible] = 1000000
```
KCGallagher, line 53:
```suggestion
        statuses[pe.property.InfectionStatus.Susceptible] = large_num
```
Should this work?

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -42,52 +44,102 @@ def __call__(self, time: float):
...
+                        # between infector/infectee is too large.
+                        self.do_infection_event(infector, infectee, timestep)
+                        number_to_infect -= 1
+
+            else:
+                # Chooses a list of cells (with replacement) for each infection
+                # event to occur in. Specifically inter-cell infections
+                # so can't be the same cell
+
+                distance_weights = [1/DistanceFunctions.dist(
```
rcw5890, line 101: Could be worth handling the case where distance=0 to avoid divide by zero

Elizabeth-Hayman: Implemented a max_weight if we get divide by zero. Just in the code for now, would suggest it as a global param, but feel I'm making a lot of those atm

rcw5890:
@Elizabeth-Hayman Thanks for those changes !
I'm wondering about your comment on how big the max_weight should be, and how we should allow some cells to be clumped on top of each other.
What do you think of setting the weight for cells on top of each other to be equal to the maximum of the valid inverse distances ? It's still slightly arbitrary, but less so than just picking 1000 !
You'd still have to have some fixed value in case all cells in the population were on top of each other, but that's such a weird situation that the value wouldn't matter much.

Elizabeth-Hayman:
Richard, I've put in some configuration to the distance weights vector, as you suggested. If all cells are on top of each each the weights are all equal which makes sense (and could be feasible if we don't have any idea of spatial complexity on a run). Annoyingly the nanmax function returns a runtime error if the argument is a vector of all nan (which would happen if all cells are on top of each other).

Elizabeth-Hayman:
Used different function to remove the runtimewarning

rcw5890 (approved):

I-Bouros (approved):

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -52,31 +64,99 @@ def test__call__(self, mock_inf, mock_force, mock_poisson):
...
+        self.pop.add_cells(1)
+        self.third_cell = self.pop.cells[2]
+        self.third_cell.add_microcells(1)
+        self.third_cell.microcells[0].add_people(1)
+        self.cell_susc.person_queue = Queue()
+        test_sweep.bind_population(self.pop)
+        test_sweep(time)
+        enqueued_people = (self.cell_susc.person_queue.qsize()
+                           + self.third_cell.person_queue.qsize())
+        self.assertEqual(enqueued_people, 0)
```
KCGallagher, line 136: Maybe we should test that this is giving the weightings we actually expect?

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -42,52 +45,117 @@ def __call__(self, time: float):
...
+                # to the maximum of all other weights.
+                # Possibly want to do twice this.
+                number_of_nans = sum(np.isnan(distance_weights))
+                if number_of_nans == len(distance_weights):
+                    distance_weights = [1 for _ in distance_weights]
+                elif number_of_nans > 0:
+                    max_weight = np.nanmax(distance_weights)
+                    distance_weights = [max_weight if np.isnan(weight)
+                                        else weight for weight in
+                                        distance_weights]
```
KCGallagher, line 120:
```suggestion
                    distance_weights = np.nan_to_num(distance_weights, nan=max_weight)
```

Much cleaner (&faster), I have not ran tests on this (and coding without autocomplete etc is surprisingly hard) but should work

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
Elizabeth-Hayman: Finding it really hard to test these explicitly, as the only thing that can really be tracked is how the queue size change, and excessive amount will obviously not help that.

Elizabeth-Hayman: Mocked up the np.nan_to_num function and checked it was called in the right places/right arguments

pyEpiabm/pyEpiabm/core/_compartment_counter.py
```diff
@@ -47,9 +38,23 @@ def report(self, old_status: InfectionStatus,
         :param new_status: Person's new infection state
         :type new_status: InfectionStatus
         """
+        if self._compartments[old_status] <= 0:
+            raise ValueError("No people of this status in this cell.")
         self._compartments[old_status] -= 1
         self._compartments[new_status] += 1
 
+    def increment_compartment(self, n_persons: int,
```
KCGallagher, line 46: Maybe should be private?

KCGallagher (approved):

PR #60, opened 2022-02-16
Fixes to cpp spatial sweeps

NicholasFan235:


KCGallagher (approved):
these look great, thanks for the correction with the sampling and pointers!

PR #61, opened 2022-02-16
Py improved gifs

### LINKED ISSUE #59: Quality of gifs
See if the color quality of the animations can be improved

KCGallagher:
# Improved gif quality
## _A half-done job_

The supplies gifs of improved quality, if the png construction method is used. I can't find a way to configure it otherwise using the given writer. `.gif` and `.mp4` seem to be the most commonly available formats and pillowwritter doesn't support mp4 (which would likely produce a better quality animation). The common 'ffmpeg' does, but I can't get the installation of this working on my laptop. I don't know if its a library that others have issues with too - otherwise maybe it would be fine to just use this?

## Root of problem

The GIF format represents color using an 8-bit palette. The maximum number of distinct colors in a GIF frame is 256. One palette entry is often spent on "clear" for animation, reducing to 255 distinct colors. When you added the PNG image, PIL converted it to use the closest available colors in the palette, but apparently nothing in the palette was that close.

To fix the colors, it is possible to convert the base image to RGB (`.convert('RGB')`) before adding, and then allow optimisation of the output gif writer.That way, PIL will generate a new palette that better approximates all colors in the combined image (in effect priorising reds in our image, which has different shades of one spectrum but no other spectra.

Closes #59


rcw5890 (approved):
Thanks for looking at this !
I think I have used ffmpeg sometime, but let's not worry about depending on it until we need a mp4

PR #62, opened 2022-02-17
Raise missing households error (Issue #26)

### LINKED ISSUE #26: Household Generation in Toy Population
## Issue

Currently we define the `make_pop()` method in the `ToyPopulationFactory` as so:

> `def make_pop(self, population_size: int, cell_number: int,  `
> `microcell_number: int, household_number: int,  `
> `if_households: bool = False):  `

However, this behaviour is undesirable. If the user doesn't define `if_households` then this will default to false, however the user will still have to define `household_num` to avoid errors, even though this is never used. Furthermore, there is no obvious/ intuitive default value to use here, and any use of a `null` value seems non ideal in the case the user specifies `if_household = true` but doesn't specify `household_number`. As this is a very public-facing function that will be referenced in user scripts/workflows/simulator scripts, I think this user behaviour should be more stable.

## Proposed Solution

These issues stem from the fact that this one function is trying to do too much, and take too many arguments. There is a separate `add_households` method, however that is called from this function. It would likely be preferable to separate these methods out, to reduce the number of arguments each one takes.

## Further Changes

We should also restructure the  `if_households = false` case, as this currently creates one household for each individual, however this is very expensive and inefficient. It would be better (and more intuitive) to have no household instances generated in this case.

This would result in errors if the household sweep was subsequently ran on this population -  this should be handled in some manner. A check within the household sweep is likely not ideal - instead a verification method could be added to the household sweep class which returns a boolean whether this sweep can be legitimately applied to this population. It is okay to raise an error here though, as the user is being stupid, rather than creating thousands of unnecessary household instances.

## Timeline

This will change how the simulator class/script works, and so will not be implemented before the initial simulation milestone is reached, to avoids delays on this step. It may well have to wait until the new year.

## References: 
Pull request #18
[Review comment in this PR](https://github.com/SABS-R3-Epidemiology/epiabm/pull/18/files#r767929894)

KCGallagher:
For info on the issue this resolves, see issue #26.

We already have a separate [function](https://github.com/SABS-R3-Epidemiology/epiabm/blob/91b775128bffadea844035b8bf6b3c44480489b4/pyEpiabm/pyEpiabm/routine/toy_population_config.py#L85) which is [optionally called](https://github.com/SABS-R3-Epidemiology/epiabm/blob/91b775128bffadea844035b8bf6b3c44480489b4/pyEpiabm/pyEpiabm/routine/toy_population_config.py#L77) if the household number is specified.

I have added a check in the household_sweep, that raises an error if households are missing. Hopefully this resolves all the issues we had with this at the end of last term.

Closes #26


Elizabeth-Hayman:
Really like this functionality, and whoever changed the if_household boolean into the if num_households>0
Both really improve the code :)

KCGallagher:
> Really like this functionality, and whoever changed the if_household boolean into the if num_households>0 Both really improve the code :)

ahaha that was me too a while ago, thanks!

pyEpiabm/pyEpiabm/sweep/household_sweep.py
```diff
@@ -35,6 +35,10 @@ def __call__(self, time: int):
                 if not infector.is_infectious():
                     continue
 
+                if infector.household is None:
+                    raise AttributeError(f"{infector} is not part of a"
```
rcw5890, line 39:
```suggestion
                    raise AttributeError(f"{infector} is not part of a "
```

rcw5890 (changes requested):

rcw5890 (approved):
Thanks

PR #72, opened 2022-03-01
Py logger

### LINKED ISSUE #39: pyEpiabm - error logging
Create error logger for python analogous to cEpiabm one. 
There are likely built in python libraries to help with this.
Also go through all TODO comments that request error logging

KCGallagher:
* Adds logging functionality
* Introduces testing class with automatic mocking of logging to suppress terminal output
* Catch and log errors in top level functions (i.e. those called directly by `simulation_flow.py`)

Closes #39

### DIFF
--- a/pyEpiabm/pyEpiabm/routine/simulation.py
+++ b/pyEpiabm/pyEpiabm/routine/simulation.py
@@ -4,6 +4,7 @@
 
 import random
 import os
+import logging
 import typing
 import numpy as np
 from tqdm import tqdm
@@ -84,6 +85,7 @@ class Simulation:
                               file_params["output_dir"])
 
         filename = os.path.join(folder, file_params["output_file"])
+        logging.error(f"Set output location to {filename}")
 
         output_titles = ["time"] + [s for s in InfectionStatus]
         if self.spatial_output:
@@ -117,6 +119,7 @@ class Simulation:
             for sweep in self.sweeps:
                 sweep(t)
             self.write_to_file(t)
+            logging.debug(f'Iteration {t} completed')
 
     def write_to_file(self, time):
         """Records the count number of a given list of infection statuses
@@ -154,3 +157,4 @@ class Simulation:
         """
         random.seed(seed)
         np.random.seed(seed)
+        logging.info(f"Set random seed to: {seed}")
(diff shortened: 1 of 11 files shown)


pyEpiabm/pyEpiabm/routine/simulation.py
```diff
@@ -154,3 +170,4 @@ def set_random_seed(seed):
         """
         random.seed(seed)
         np.random.seed(seed)
+        logging.info(f"Set random seed to: {seed}")
```
rcw5890, line 173: Any reason for doing this here, but not the other times random seed is set ?

rcw5890 (changes requested):
And here's a copy of that code we were just looking at:
```
def log_exceptions(message):
    def decorator(func):
        def wrapper(*args, **kwargs):
            try:
                func(*args, **kwargs)
            except Exception as e:
                logging.exception(f"{type(e).__name__}: {message}")
        return wrapper
    return decorator
```

```
class A:
    
    @log_exceptions('Fatal error in method')
    def method(self, x):
        x = 1/0
        return x
    
a=A()
a.method(1)
```

pyEpiabm/pyEpiabm/routine/simulation.py
KCGallagher: I've added it to the population seeds too

pyEpiabm/pyEpiabm/utility/exception_logger.py
```diff
@@ -0,0 +1,24 @@
...
+    :param message: Optional message to append to default log text
+    :type message: str
+    """
+    def decorator(func):
+        def wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except Exception as e:
+                logging.exception(f"{type(e).__name__} in "
+                                  + f"{func.__qualname__}()" + message)
```
rcw5890, line 22:
```suggestion
                                  + f"{func.__qualname__}(): " + message)
```
So that you dont have to worry about this in your message

rcw5890 (changes requested):

pyEpiabm/pyEpiabm/utility/exception_logger.py
rcw5890: Ok lets not do that

rcw5890 (approved):

PR #78, opened 2022-03-02
distance restriction in spatial sweep

Elizabeth-Hayman:
Very minor addition to the spatial sweep, to allow a maximum infection distance to be set in both the Covidsim and non-covidsim cases

### DIFF
--- a/pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
+++ b/pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
@@ -88,7 +88,8 @@ class SpatialSweep(AbstractSweep):
                     infection_distance = DistanceFunctions.dist(
                      cell.location, infectee_cell.location) / Parameters.\
                         instance().infection_radius
-                    if (infection_distance < random.random()):
+                    r = random.random()
+                    if (infection_distance < r):
                         # Covidsim rejects the infection event if the distance
                         # between infector/infectee is too large.
                         self.do_infection_event(infector, infectee, timestep)
@@ -118,13 +119,16 @@ class SpatialSweep(AbstractSweep):
                     max_weight = np.nanmax(distance_weights)
                     distance_weights = np.nan_to_num(distance_weights,
                                                      nan=max_weight)
+                # Use of the cutoff distance idea from CovidSim.
+                cutoff = Parameters.instance().infection_radius
+                distance_weights = [weight/sum(distance_weights) for weight in
+                                    distance_weights if weight > 1/cutoff]
                 cell_list = random.choices(possible_infectee_cells,
                                            weights=distance_weights,
                                            k=number_to_infect)
                 # Each infection event corresponds to a infectee cell
                 # on the cell list
                 for infectee_cell in cell_list:
-
                     # Sample at random from the infectee cell to find
                     # an infectee
                     infectee = random.sample(infectee_cell.persons, 1)[0]


pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -88,7 +88,8 @@ def __call__(self, time: float):
                     infection_distance = DistanceFunctions.dist(
                      cell.location, infectee_cell.location) / Parameters.\
                         instance().infection_radius
-                    if (infection_distance < random.random()):
+                    r = random.random()
+                    if (infection_distance < r):
```
KCGallagher, line 92: We can discard this change now - it was only for printing!

rcw5890 (approved):

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -118,13 +118,16 @@ def __call__(self, time: float):
                     max_weight = np.nanmax(distance_weights)
                     distance_weights = np.nan_to_num(distance_weights,
                                                      nan=max_weight)
+                # Use of the cutoff distance idea from CovidSim.
+                cutoff = Parameters.instance().infection_radius
+                distance_weights = [weight if (weight > 1/cutoff) else 0 for
+                                    weight in distance_weights]
```
KCGallagher, line 124: I though you had added checks that this list is of the right length (and has non-zero elements) but can't see them here?

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -64,6 +67,7 @@ def test__call__(self, mock_inf, mock_force, mock_poisson, mock_dist,
         self.microcell_susc.add_people(1)
```
KCGallagher, line 67: Adding the `microcell_susc` to `self.pop` could be done in `setUpClass` if you want to use it in multiple tests - best not to change anything in self in the tests as unittest doesn't guarantee it will run the tests in the order here! 

In general, probably best to avoid modifying anything in self inside a test - you can access attributes of self, but copy them if you want to modify them! Unit tests should run separately, and independent of each other (whereas the second test in this file requires the first to have modified self)

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -45,7 +48,7 @@ def test__call__(self, mock_inf, mock_force, mock_poisson, mock_dist,
         mock_inf.return_value = 10.0
         mock_force.return_value = 100.0
         time = 1
-        Parameters.instance().infection_radius = 5000
+        Parameters.instance().infection_radius = 1000
```
KCGallagher, line 51: Do we have a check that no infections occur when this is really small?

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
Elizabeth-Hayman: But one of the tests needs to catch the case when the population only has one cell.

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
Elizabeth-Hayman: Yep, it's like the penultimate bit

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
Elizabeth-Hayman: Do now :)

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -118,13 +123,30 @@ def __call__(self, time: float):
...
+                    # Will catch a list of zeros
+                    1/sum(distance_weights)
+                    cell_list = random.choices(possible_infectee_cells,
+                                               weights=distance_weights,
+                                               k=number_to_infect)
+                except ZeroDivisionError:
+                    logging.exception(f"{type(ValueError).__name__}: no cells"
+                                      + f" within radius {cutoff} of"
+                                      + f" cell {cell.id} at location"
+                                      + f" {cell.location}. Skip cell.")
```
KCGallagher, line 144:
```suggestion
                                      + f" {cell.location} - skipping cell.")
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -45,77 +48,91 @@ def test__call__(self, mock_inf, mock_force, mock_poisson, mock_dist,
...
         mock_nan.return_value = [2, 2]
-
-        self.infectee.update_status(InfectionStatus.Susceptible)
         mock_dist.side_effect = [0, 2]
-        self.pop.add_cells(1)
-        self.third_cell = self.pop.cells[2]
+        test_pop.add_cells(1)
+        self.third_cell = test_pop.cells[2]
         self.third_cell.add_microcells(1)
         self.third_cell.microcells[0].add_people(1)
```
KCGallagher, line 130:
```suggestion
        third_cell = test_pop.cells[2]
        third_cell.add_microcells(1)
        third_cell.microcells[0].add_people(1)
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -45,77 +48,91 @@ def test__call__(self, mock_inf, mock_force, mock_poisson, mock_dist,
         mock_inf.return_value = 10.0
         mock_force.return_value = 100.0
         time = 1
-        Parameters.instance().infection_radius = 5000
+        Parameters.instance().infection_radius = 1000
 
+        test_pop = self.pop
```
KCGallagher, line 53: @rcw5890 do we need a deep copy of the population here for stability, or should this suffice?

EDIT: They give the same hash for me (i.e. point to the same object) so I believe this is not sufficient and you will need a deep copy - otherwise changes to test_pop will also affect self.pop. Richard might be able to advise on best practice here though

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -42,6 +43,10 @@ def __call__(self, time: float):
         # cells is more than one (edge case but worth having)
         if len(self._population.cells) == 1:
             return
+        # If infecion radius is set to zero no infections will occur so
```
rcw5890, line 46:
```suggestion
        # If infection radius is set to zero no infections will occur so
```

rcw5890 (changes requested):

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
rcw5890: @Elizabeth-Hayman @KCGallagher I agree we ought to get a new copy of the population. Unfortunately, copy.deepcopy causes weird problems possibly related to logging, I'm guessing you saw the same thing at some point ?

The better option may be to convert `setUpClass` (called once before running all tests in the class) to `setUp` (called before each test). This will give us a fresh population for each test.

As written `test_do_infection_event` seems to depend on some of the changes to the population that happen in the previous test, so that will require some minor changes !

Elizabeth-Hayman: I think I understand this and have made the relevant changes

KCGallagher: Ahh that sounds the ideal solution (and may even be worth using elsewhere in the code too) - I didn't realise there was this difference between `setUp` and `setUpClass`. We won't need the `super` line in the setup now as we're not overwriting the `setUpClass` from my custom parent, but it will be worth confirming that the logger doesn't print (I haven't tested this scenario but it should still work)

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -118,13 +123,31 @@ def __call__(self, time: float):
...
-                                           weights=distance_weights,
-                                           k=number_to_infect)
+                # Use of the cutoff distance idea from CovidSim.
+                cutoff = Parameters.instance().infection_radius
+                distance_weights = [weight if (cutoff > 1/weight) else 0 for
+                                    weight in distance_weights]
+                # Will catch the case if distance weights isn't configured
+                # correctly and returns the wrong length.
+                assert len(distance_weights) == len(possible_infectee_cells), (
+                    "distance weights are not the same length as cell list")
```
KCGallagher, line 133:
```suggestion
                    "Distance weights are not the same length as cell list")
```

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -118,13 +123,31 @@ def __call__(self, time: float):
...
+
+                try:
+                    # Will catch a list of zeros
+                    1/sum(distance_weights)
```
KCGallagher, line 137:
```suggestion
                    1 / sum(distance_weights)
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -17,25 +18,29 @@ def setUpClass(cls) -> None:
         """Initialises a population with one cell and one person in
         the cell.
         """
-        cls.pop = Population()
-        cls.pop.add_cells(1)
-        cls.cell_inf = cls.pop.cells[0]
+        super(TestSpatialSweep, cls).setUpClass()
```
KCGallagher, line 21: Does it not work without this line? I thought it might but no worries if it doesn't (just useful for me to know)

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -45,77 +50,96 @@ def test__call__(self, mock_inf, mock_force, mock_poisson, mock_dist,
...
 
         # All (one) valid distances also won't call nan_to_num
         mock_dist.return_value = 2
         Parameters.instance().do_CovidSim = True
-        self.cell_susc.person_queue = Queue()
-        test_sweep.bind_population(self.pop)
+        cell_susc.person_queue = Queue()
+        test_sweep.bind_population(test_pop)
         test_sweep(time)
         self.assertFalse(mock_nan.called)
```
KCGallagher, line 107:
```suggestion
        mock_nan.assert_not_called()
```
I believe this is more stable?

KCGallagher (commented):
Looks good - couple of tiny changes!

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
Elizabeth-Hayman: It works with no set up class at all (which makes sense). Initially I just removed that one line and then it failed bc I guess it was overwriting the inheritance

Elizabeth-Hayman:
logger doesn't appear to be printing with the mock :)

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -118,13 +123,31 @@ def __call__(self, time: float):
...
+                distance_weights = [weight if (cutoff > 1/weight) else 0 for
+                                    weight in distance_weights]
+                # Will catch the case if distance weights isn't configured
+                # correctly and returns the wrong length.
+                assert len(distance_weights) == len(possible_infectee_cells), (
+                    "Distance weights are not the same length as cell list")
+
+                try:
+                    # Will catch a list of zeros
+                    1 / sum(distance_weights)
```
KCGallagher, line 137: Are we using this result? If not just check the sum of the list is non-zero (copy the source code from the random module)

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -118,13 +123,31 @@ def __call__(self, time: float):
                     max_weight = np.nanmax(distance_weights)
                     distance_weights = np.nan_to_num(distance_weights,
                                                      nan=max_weight)
-                cell_list = random.choices(possible_infectee_cells,
-                                           weights=distance_weights,
-                                           k=number_to_infect)
+                # Use of the cutoff distance idea from CovidSim.
+                cutoff = Parameters.instance().infection_radius
+                distance_weights = [weight if (cutoff > 1/weight) else 0 for
+                                    weight in distance_weights]
```
KCGallagher, line 129:
```suggestion
                distance_weights = [weight if (cutoff > 1/weight) else 0
                                    for weight in distance_weights]
```
I believe this is slightly better practice for the line split?

pyEpiabm/pyEpiabm/tests/test_sweep/test_spatial_sweep.py
```diff
@@ -45,77 +43,96 @@ def test__call__(self, mock_inf, mock_force, mock_poisson, mock_dist,
...
+
+        test_sweep.bind_population(test_pop)
         test_sweep(time)
-        self.assertTrue(self.cell_susc.person_queue.empty())
+        self.assertTrue(cell_susc.person_queue.empty())
 
         # Change infector's status to infected
         self.infector.update_status(InfectionStatus.InfectMild)
-        test_sweep.bind_population(self.pop)
+        test_sweep.bind_population(test_pop)
```
KCGallagher, line 74: I am interested, do we _need_ to recall this method each time - surely the population is already bound and we modify the population inplace? The tests seem to work fine without it so I was confused why you've retyped it each time (there may well be a good reason so do let me know if so!)

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
Elizabeth-Hayman: I agree the source code seems to produce ValueErrors as described, but not able to get any to raise on my machine

Elizabeth-Hayman: Written it in manually but not the slickest

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -33,15 +34,22 @@ def __call__(self, time: float):
         infection event occurs on individual and cell infectiousness
         and susceptibility.
 
-        :param time: Current simulation time
-        :type time: int
+        Parameters
+        ----------
+        time : float
+            Current simulation time
+
```
KCGallagher, line 41: I was updating docs, and noticed you are handling time as an int, but it should be a float to be compatible with Luke' changes

KCGallagher (changes requested):
Time should now be a `float` after @lukedtc 's changes, but still an `int` here. Also the call function is super long - can we split this up into smaller functions (such as separate functions/private methods for the if CovidSim clauses)? Python functions are typically 20 lines max and its over 100

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
Elizabeth-Hayman: Okey cool, is that all sorted?

KCGallagher: The docs are, the code still generates/uses time as an int.

rcw5890 (approved):

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -71,77 +79,149 @@ def __call__(self, time: float):
...
+        possible_infectee_cells : List
+            List of possible cells to infect
+        number_to_infect : int
+            maximum number of people to infect
+
+        Returns
+        ----------
+        infectee_list : List
+            List of exposed people to test an infection event
+        """
```
KCGallagher, line 112:
```suggestion
            List of exposed people to test an infection event

        """
```

pyEpiabm/pyEpiabm/sweep/spatial_sweep.py
```diff
@@ -71,77 +79,149 @@ def __call__(self, time: float):
...
+
+        Parameters
+        ----------
+        infector : Person
+            Infector instance of person
+        possible_infectee_cells : List
+            List of possible cells to infect
+        number_to_infect : int
+            maximum number of people to infect
+        """
```
KCGallagher, line 181:
```suggestion
            Maximum number of people to infect

        """
```

KCGallagher (approved):

PR #82, opened 2022-03-04
Py setup tests

KCGallagher:
In response to @rcw5890 's [suggestion](https://github.com/SABS-R3-Epidemiology/epiabm/pull/78#discussion_r819082628), I have implemented this in our other tests to ensure there is no dependence on the order tests are run within each class/file. In some cases this could increase the run time slightly, but in no cases more than 10% and the difference is often negligible. Let me know what you think!


rcw5890 (changes requested):
It looks great, however `setup` is not a classmethod, and it should take self as its argument.

rcw5890 (approved):

PR #84, opened 2022-03-04
reconfigure of place class

Elizabeth-Hayman:
As part of the reconfigure of the place sweep I have put in a few new ideas from Covidsim. NONE OF THIS IS TESTED.
Wanted to see what people thought before tidying it.

- places are stratified by different groups, which may be fixed or vary on each update status. I have implemented this, and started thinking about how this would be initialised.
- different group structures for different place types. I have documented some ideas about basic structures in the enum PlaceTypes.
- In Covidsim only people in the same place AND group can infect each other. Richard agrees this seem pointless if the populations are fixed, may as well have multiple places. Another opportunity for the doCovidsim flag.
- Changes to the update place sweep would very much lend themselves to reading in a parameters file as we were discussing


Elizabeth-Hayman:
Further Covidsim comment: people exist in precisely one location of each type (at each timestep?). Assigned either by them being the fixed population, or weighted by distance to household.

pyEpiabm/pyEpiabm/core/person.py
```diff
@@ -98,14 +98,21 @@ def update_time_to_status_change(self) -> None:
         new_time = random.randint(1, 10)
         self.time_of_status_change = new_time
 
-    def add_place(self, place):
+    def add_place(self, place, person_group: int = 0):
         """Method adds a place to the place list if the person visits
-        or is associated with this place.
+        or is associated with this place. Places are saved as a tuple
+        with the place as the first entry and group person is associated
```
KCGallagher, line 104: "and _the_ group _a_ person is associated with" perhaps? Not sure I understand this docstring

pyEpiabm/pyEpiabm/core/person.py
```diff
@@ -98,14 +98,21 @@ def update_time_to_status_change(self) -> None:
...
+        with as the second.
+
+        :param place: Place person should be removed from
```
KCGallagher, line 107:
```suggestion
        :param place: Place person should be added to
```

pyEpiabm/pyEpiabm/core/person.py
```diff
@@ -114,7 +121,9 @@ def remove_place(self, place):
...
         :type place: Place
         """
-        if place not in self.places:
+        place_list = [i[0] for i in self.places]
+        if place not in place_list:
             raise KeyError("Person not found in this place")
         else:
-            self.places.remove(place)
+            ind = place_list.index(place)
+            self.places.pop(ind)
```
KCGallagher, line 129: If all people from a group are removed from a place, will we need to reassigned the group index?

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -69,14 +74,21 @@ def set_susceptibility(self, susceptibility: float):
         """
         self.susceptibility = susceptibility
 
-    def add_person(self, person: Person):
+    def add_person(self, person: Person, person_group: int = 0):
         """Add a person into the place.
 
         :param person: Person to add
         :type person: Person
+        :param person_group: key for the person group dictionary
```
KCGallagher, line 82:
```suggestion
        :param person_group: Key for the person group dictionary
```

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -87,13 +99,35 @@ def remove_person(self, person: Person):
...
+            # key lmk.
             person.remove_place(self)
             self.persons.remove(person)
 
-    def empty_place(self):
-        """Remove all people from place. For example
+    def get_group_index(self, person):
+        """Get the group of a person in the place.
+
+        :param person: Person to remove from place
```
KCGallagher, line 113:
```suggestion
        :param person: Person associated with group
```

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -87,13 +99,35 @@ def remove_person(self, person: Person):
         if person not in self.persons:
             raise KeyError("Person not found in this place")
         else:
+            group_index = self.get_group_index(person)
+            self.person_groups[group_index].remove(person)
+            # If anyone can think of a better way to remove
+            # people from a dictionary without knowing the
+            # key lmk.
```
KCGallagher, line 106: The key is `group_index` isn't it? This looks sensible!

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -87,13 +99,35 @@ def remove_person(self, person: Person):
...
+    def empty_place(self, person_groups: list = [0]):
+        """Remove all people from place who are in a specific
+        person group. For example
         a restaurant or park might regularly change
-        all occupants each timestep.
+        all occupants each timestep, but workers at the
+        restaurant will be present each timestep.
+
+        :param person_groups: list of person_group
+        indicies to be removed
```
KCGallagher, line 128:
```suggestion
        :param person_groups: List of person_group
            indicies to be removed
```

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -20,25 +20,90 @@ def __call__(self, time: int):
...
-                        place.add_person(place.cell.persons[i-1])
-                        count += 1
+                if place.place_type.value == 1:  # HOTEL
+                    place.empty_place()
+                    self.update_place_group(place)
+                    # not sure how to handle this yet so
+                    # just going to completely update each time
+                    # as in the previous code
+                elif place.place_type.value == 1:  # CAREHOME
+                    if not place.initialised:
```
KCGallagher, line 36: can we have a separate initialise function maybe? Seems a waste to be running this check on every timestep, and presumably we won't be adding new places mid-simulation? Not a big cost in the grand scheme of things though

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -20,25 +20,90 @@ def __call__(self, time: int):
...
+                        # Initialise the fixed population of workers
+                        max_cap = 10  # for example, would actually come in
+                        # from a dictionary
```
KCGallagher, line 39: How does CovidSim do it? This seems sensible, or could easily generate from normal distribution which would probably also be as realistic as needed (but I know we do want to copy CovidSim)

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -20,25 +20,90 @@ def __call__(self, time: int):
...
+                elif place.place_type.value == 3:  # RESTAURANT
+                    if not place.initialised:
+                        # Initialise the fixed population
+                        max_cap = 10
+                        self.update_place_group(place, max_capacity=max_cap,
+                                                group_index=0)
+                        place.initialised = True
+                    # Variable population is people not in the fixed pop.
+                    # Changed at each timestep
+                    place.empty_place(1)
```
KCGallagher, line 56:
```suggestion
                    place.empty_place(person_groups=[1])
```
Readability suggestion, as its really not clear what the argument is otherwise from the function name. Also think you need to pass a list not a number?

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -20,25 +20,90 @@ def __call__(self, time: int):
...
+                    person_list = [person for person in place.cell.persons if
```
KCGallagher, line 57: Will this not add the same people back in? Do we not need to change the people in the place between L56 and L57? I am probably missing something here though!

EDIT: Apologies - I see now that we randomly select from the person_list  - so they won't all make it in. I wonder whether we change the name of that variable to `possible_person_list` or `candidate_list` or something similar, so it doesn't look like that all people in that list will be added to the place?

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -20,25 +20,90 @@ def __call__(self, time: int):
...
+        """Specific method to update people in a place or place group.
+
+        :param place: Place to change
+        :type place: Place
+        :param max_capacity: maximum people of this group in this place
+        :type max_capacity: int
+        :param group_index: key for the person group dictionary
+        :type group_index: int
+        :param person_list: list of people that may be present in the cell
+        :type person_list: list
```
KCGallagher, line 92:
```suggestion
        :param max_capacity: Maximum people of this group in this place
        :type max_capacity: int
        :param group_index: Key for the person group dictionary
        :type group_index: int
        :param person_list: List of people that may be present in the cell
        :type person_list: list
```

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -20,25 +20,90 @@ def __call__(self, time: int):
...
+        """
+        # If a specific list of people is not provided, use the whole cell
+        if person_list is None:
+            person_list = place.cell.persons
+
+        # Ensure that the number of people put in the place
+        # is at most its capacity or the total number of
+        # people in the cell.
+        max_capacity = np.minimum(max_capacity, len(person_list))
+        new_capacity = random.randint(1, max_capacity)
```
KCGallagher, line 102: Out of interest, does CovidSim use a uniform distribution? Not saying we have to replicate their version if it is very complicated though!

KCGallagher (changes requested):
I like this - it adds a lot of complexity to the simulation in a simple way so we could get more complex behaviour. Maybe some of the max capacities can be read in from the parameters file, to make it more configurable?

pyEpiabm/pyEpiabm/core/person.py
Elizabeth-Hayman: Not sure. The way I've suggested in the PlaceType enum has each index vaguely meaning something (workers, visitors etc) but that was only a suggestion

pyEpiabm/pyEpiabm/core/place.py
Elizabeth-Hayman: Ah that's a left over doc string

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
Elizabeth-Hayman: Yea, started writing a separate place_initialise sweep but wasn't sure. Would make sense though so will do that.

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
Elizabeth-Hayman: Yes - Covidsim also makes sure people aren't assigned to more than one group/place of each type so would decrease this list as people are added in

pyEpiabm/pyEpiabm/core/person.py
KCGallagher: Ahh yes that makes sense!

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
Elizabeth-Hayman: Lines ~2090 in SetUpModel seem to set the place size to be either lognormal or poisson added to the previous number of people.
So can definitely implement that

Elizabeth-Hayman: For some reason place size seems to be saved in a variable called treat_end_time

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
Elizabeth-Hayman: see discussion above

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
Elizabeth-Hayman: Then they have parameters for max size, min size, mean size, an exponent's in there too. The defaults are verrry generous, like range [1, 10^20] , and lots of zeros

KCGallagher: Ahh of course, completely logical there. A lognormal dist sounds sensible though, and hopefully gives significantly more realism than the constant limit without adding too much complexity to implement

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -33,10 +35,14 @@ def __init__(self, loc: typing.Tuple[float, float],
         """
         self._location = loc
         self.persons = []
+        self.person_groups = {0: []}
+        self.num_person_groups = 0
```
KCGallagher, line 39: Is the default here 1 if you have a group 0 in L38?

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -33,10 +35,14 @@ def __init__(self, loc: typing.Tuple[float, float],
...
         self.place_type = place_type
         self.max_capacity = 50
+        self.mean_capacity = 25
```
KCGallagher, line 42: Can this be specified in params?

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -59,6 +65,15 @@ def set_max_cap(self, max_capacity: int):
         """
         self.max_capacity = max_capacity
 
+    def set_mean_cap(self, mean_capacity: float):
+        """Sets the maximum capacity of a place.
+
+        :param mean_capacity: Maximum number of people
+            allowed in place
+        :type mean_capacity: float
```
KCGallagher, line 73:
```suggestion
        :param mean_capacity: Mean capacity of place type
        :type mean_capacity: float
```
Not sure if this matches its use, but do correct the docstring either way

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -81,17 +96,24 @@ def set_susceptibility(self, susceptibility: float):
...
 
-    def add_person(self, person: Person):
+    def add_person(self, person: Person, person_group: int = 0):
         """Add a person into the place.
 
         Parameters
         ----------
         person: Person
             Person to add
+        person_group : Int
```
KCGallagher, line 106:
```suggestion
        person_group : int
```

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -105,14 +127,35 @@ def remove_person(self, person: Person):
...
-        """Remove all people from place. For example
+    def get_group_index(self, person):
+        """Get the group of a person in the place.
+
+        :param person: Person associated with group
+        :type person: Person
+        """
+        place_list = [i[0] for i in person.places]
+        ind = place_list.index(self)
+        return person.places[ind][1]
```
KCGallagher, line 143: Why not just search the `person_group` attribute of self? I'm confused why the place is searching though the person's list of places, rather than their own list of people, but I might be missing something here

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,64 @@
...
+
+        :param time: Current simulation time
+        :type time: int
+        """
+
+        # Double loop over the whole population, clearing places
+        # of the variable population and refilling them.
+
+        # May want a param dict to be called in from file etc.
+        # place_params = Parameters.instance().place_params
```
KCGallagher, line 25: Yes I think this would be preferable

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/core/place.py
Elizabeth-Hayman: Was asking Nick if we could set some function on the placetype enum to default parameters for each enum, but by that point probably just want to be reading in stuff. Possibly want all the parameters to be a list length 5 so might need some indexing

pyEpiabm/pyEpiabm/core/place.py
Elizabeth-Hayman: I think I did have it that way originally, but my implementation of searching a dictionary was longer than this. Have put in a check to ensure person is in place though.

Elizabeth-Hayman:
Hotel place types only exist if airports are activated. In both the US and UK this is set or defaulted to 0 so won't happen. If hotels do exist they are put last in the array of places.

Carehomes have a lot of default parameters preset, including setting the placetype to -1 which will mean it is not actually called into the code.

So if we follow directly the UK parameter data then (as the supplementary material suggests) we wouldn't have these, just the school/workplace variables

pyEpiabm/pyEpiabm/core/place.py
KCGallagher: Yeah I think reading in stuff is the way to go - at the moment I think we're sticking everything we'll want to read in in the parameters file (so it could be theoretically configurable from there), but will move that to a `.json` or similar soon. Given the rigis enum setup I think lists are fine - just worth making sure we have the place types we want (i.e. no schools for example) before we commit to that!

KCGallagher:
> Hotel place types only exist if airports are activated. In both the US and UK this is set or defaulted to 0 so won't happen. If hotels do exist they are put last in the array of places.
> 
> Carehomes have a lot of default parameters preset, including setting the placetype to -1 which will mean it is not actually called into the code.
> 
> So if we follow directly the UK parameter data then (as the supplementary material suggests) we wouldn't have these, just the school/workplace variables

Does that mean we will want to change the enums to include schools but not hotels for example? I think it would be fair enough to leave hotels out as their paper says air travel is less important in the UK anyway, it is more important in the US but we'd never have simulations that big in the python. I am looking into carehomes for my next PR anyway so will let you know if I want a place enum for those!

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -105,14 +126,38 @@ def remove_person(self, person: Person):
...
+        :type person: Person
+        """
+        if person not in self.persons:
+            raise KeyError("Person not found in this place")
+        else:
+            place_list = [i[0] for i in person.places]
+            ind = place_list.index(self)
+        return person.places[ind][1]
+
+    def empty_place(self, groups_to_empty: list = [0]):
```
rcw5890, line 147: I'm not convinced that this is a good default behavior. I would either require that a group be passed, or make the default be to empty all groups.

rcw5890 (changes requested):

pyEpiabm/pyEpiabm/core/place.py
Elizabeth-Hayman: Put my parameters in as a dictionary anyway

pyEpiabm/pyEpiabm/core/place.py
Elizabeth-Hayman: Defaulted to empty groups

Elizabeth-Hayman:
Yes, I've changed the enum up a little so can read in Covidsim parameters and apply them to the correct place.

rcw5890 (approved):

pyEpiabm/pyEpiabm/tests/test_sweep/test_initialise_place_sweep.py
```diff
@@ -0,0 +1,90 @@
...
+        place = test_pop.cells[0].places[0]
+        place.place_type = PlaceType.Workplace
+        person = test_pop.cells[0].persons[0]
+        mock_update.return_value = None
+        mock_weights.return_value = [[person], [1]]
+
+        test_sweep = pe.sweep.InitialisePlaceSweep()
+        test_sweep.bind_population(test_pop)
+        test_sweep()
+        mock_update.called_with(place, [person], [1], 14.28, 5927, 0)
```
KCGallagher, line 58: I have never seen this before, are you sure you don't mean `assert_called_with()`?

pyEpiabm/pyEpiabm/tests/test_sweep/test_update_place_sweep.py
```diff
@@ -1,54 +1,93 @@
...
+        self.cell = self.pop.cells[0]
+        self.pop.cells[0].add_microcells(1)
+        self.microcell = self.cell.microcells[0]
+        self.pop.cells[0].microcells[0].add_people(1)
+        self.person = self.pop.cells[0].microcells[0].persons[0]
+        self.person.update_status(InfectionStatus.InfectMild)
+        self.microcell.add_place(1, (1, 1), PlaceType.Workplace)
+        self.place = self.cell.places[0]
+        Parameters.instance().time_steps_per_day = 1
+        self.time = 1
```
KCGallagher, line 30:
```suggestion
        self.time = 1.0
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_update_place_sweep.py
```diff
@@ -1,54 +1,93 @@
...
-        pe.Parameters.instance().time_steps_per_day = 1
-        cls.time = 1
+        self.pop = Population()
+        self.pop.add_cells(1)
+        self.cell = self.pop.cells[0]
+        self.pop.cells[0].add_microcells(1)
+        self.microcell = self.cell.microcells[0]
+        self.pop.cells[0].microcells[0].add_people(1)
+        self.person = self.pop.cells[0].microcells[0].persons[0]
+        self.person.update_status(InfectionStatus.InfectMild)
```
KCGallagher, line 26: Do you need to artificially make this person infectious too? (or infact do they need to be infectious at all?)

KCGallagher (changes requested):
How complete is this sweep? Could we add the `InitialisePlaceSweep` to the initial sweeps list and run an ex

pyEpiabm/pyEpiabm/core/place.py
```diff
@@ -59,6 +65,14 @@ def set_max_cap(self, max_capacity: int):
...
 
+    def set_mean_cap(self, mean_capacity: float):
+        """Sets the mean capacity of a place.
+
+        :param mean_capacity: Mean capacity of place type
+        :type mean_capacity: float
+        """
+        self.mean_capacity = mean_capacity
+
     def set_infectiousness(self, infectiousness: float):
```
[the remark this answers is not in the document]
Elizabeth-Hayman: I think now we don't (for mean capacity) because the relevant params are read in from file and the same across any one placetype. Good spot! 
Susceptibility may be different as might want to change that up depending on time? But you thought Covidsim susceptibility was dependent on the people not particularly the place

pyEpiabm/pyEpiabm/core/parameters.py
```diff
@@ -160,12 +162,43 @@ def __init__(self):
             self.asympt_infectiousness = 1
             self.latent_to_sympt_delay = 0.5
 
+            # PLACE PARAMS
+            # Copied straight from Covidsim so for four placetypes:
+            # Three school systems (primary, secondary, sixth form)
+            # and a workplace option, hence array size four.
+            # any single values apply to all place types
+            self.place_params = {"age_group1_min_age": [5, 11, 18, 16],
```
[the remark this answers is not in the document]
Elizabeth-Hayman: The latest edits to the place sweeps use these to configure people in places (though means if we use default age of zero nobody will go to a place!)

Promise will document today

pyEpiabm/pyEpiabm/core/parameters.py
```diff
@@ -160,12 +162,43 @@ def __init__(self):
...
+                                 "max_size": [0, 0, 0, 5927],
+                                 "size_power": [0, 0, 0, 1.34],
+                                 "size_offset": [0, 0, 0, 5.35],
+                                 "mean_num_groups": [25,  25,  100, 10]}
+
+            # Build infection state transition matrix and set as parameter
+            matrix_object = StateTransitionMatrix()
+            self.state_transition_matrix =\
+                matrix_object.create_state_transition_matrix()
+            self.place_transmission = 0.1
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Oooo remeber it causing merge issues but assumed best not to delete. Will sort.

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -24,25 +26,77 @@ def __call__(self, time: float):
...
+        # people in the cell.
+        new_capacity = np.random.lognormal(mean_capacity)
+        new_capacity = min(new_capacity, max_capacity, len(person_list))
+        if len(person_list) <= 0:
+            logging.warning("No people in the person list supplied.")
+        count = 0
+        while count < new_capacity:
+            if person_weights is not None:
+                assert len(person_weights) == len(person_list), 'Weights list'
+                'given is a different size to the person list.'
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Gotcha

pyEpiabm/pyEpiabm/tests/test_sweep/test_initialise_place_sweep.py
Elizabeth-Hayman: possibly

pyEpiabm/pyEpiabm/tests/test_sweep/test_update_place_sweep.py
Elizabeth-Hayman: apparently not

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,130 @@
...
+                                           place.person_groups[i]]
+
+                elif place.place_type.name == "CareHome":  # CAREHOME
+                    # Initialise the fixed population of workers
+                    helper.update_place_group(place, group_index=0)
+                    # Initialise a fixed population of residents
+                    person_list = [person for person in cell.persons
+                                   if person.age > 65]
+                    helper.update_place_group(place, group_index=1,
+                                              person_list=person_list)
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Outdoor spaces will be done in the last lines of this sweep where Update sweep is called. Because they only have a variable population

pyEpiabm/pyEpiabm/core/place.py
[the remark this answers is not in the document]
Elizabeth-Hayman: see new branch

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,130 @@
+#
+# Sweep to initialise people present in a place
+#
+
+import numpy as np
+
+from .abstract_sweep import AbstractSweep
+from .update_place_sweep import UpdatePlaceSweep
+from pyEpiabm.core import Parameters
+
```
KCGallagher, line 10:
```suggestion
from pyEpiabm.core import Parameters

from .abstract_sweep import AbstractSweep
from .update_place_sweep import UpdatePlaceSweep

```

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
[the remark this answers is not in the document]
KCGallagher: That makes a lot of sense (but equally I still don't think this is obvious so wouldn't mind a tiny comment explaining that for readability, as its the only place type which isn't called explicitly)

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,130 @@
...
+            Dictionary of parameters with age structure data
+
+        Returns
+        -------
+
+        person_list : list
+            List of people who may be in the place
+        weights : list
+            Corresponding weights for the person list
+        """
```
KCGallagher, line 107:
```suggestion
            Corresponding weights for the person list

        """
```

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,130 @@
...
+            Place to add people to
+        params : dict
+            Dictionary of parameters with age structure data
+
+        Returns
+        -------
+
+        person_list : list
+            List of people who may be in the place
+        weights : list
```
KCGallagher, line 105:
```suggestion
        typing.List[Person]
            List of people who may be in the place
        typing.List[float]
```
Don't tend to specify internal variable names in the return for the docs, but we can specify the typing to differentiate between them still

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,130 @@
...
+    """Class to update people in the "Place"
+    class.
+    """
+    def __call__(self):
+        """Given a population structure, updates the people
+        present in each place at a specific timepoint.
+
+        :param time: Current simulation time
+        :type time: int
+        """
```
KCGallagher, line 22:
```suggestion
        Parameters
        ----------
        time : float
            Current simulation time

        """
```

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -24,25 +27,77 @@ def __call__(self, time: float):
...
+            Place to change
+        max_capacity : int
+            Maximum people of this group in this place
+        group_index: int
+            Key for the person group dictionary
+        person_list: list
+            List of people that may be present in the cell
+        person_weights : list
+            Weights for people in list
+        """
```
KCGallagher, line 76:
```suggestion
            Weights for people in list

        """
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_update_place_sweep.py
```diff
@@ -1,54 +1,95 @@
 import unittest
+from unittest import mock
 
-import pyEpiabm as pe
+from pyEpiabm.property.place_type import PlaceType
+from pyEpiabm.core.population import Population
+from pyEpiabm.core.parameters import Parameters
+from pyEpiabm.sweep.update_place_sweep import UpdatePlaceSweep
```
KCGallagher, line 7:
```suggestion
from pyEpiabm.property import PlaceType
from pyEpiabm.core import Parameters, Population
from pyEpiabm.sweep import UpdatePlaceSweep
```
Shouldn't cause any import errors, but always best to import from subpackage rather than file!

pyEpiabm/pyEpiabm/tests/test_sweep/test_update_place_sweep.py
```diff
@@ -1,54 +1,95 @@
...
+        self.assertDictEqual(place.person_groups,
+                             {0: [], 1: [person]})
+        self.place.empty_place([1])
+        test_sweep.update_place_group(place, person_list=[person],
+                                      person_weights=[1], group_index=1)
+        self.assertDictEqual(place.person_groups,
+                             {0: [], 1: [person]})
+        self.place.empty_place()
+        test_sweep.update_place_group(place, person_list=[])
+        log_mock.called
```
KCGallagher, line 69: Line incomplete?

pyEpiabm/pyEpiabm/tests/test_sweep/test_update_place_sweep.py
```diff
@@ -1,54 +1,95 @@
...
+        place = test_pop.cells[0].places[0]
+        person = test_pop.cells[0].persons[0]
+        mock_update.return_value = None
+        test_sweep = UpdatePlaceSweep()
+        test_sweep.bind_population(test_pop)
+        self.assertFalse(place.persons)
+
+        place.place_type = PlaceType.OutdoorSpace
+        test_sweep(1)
+        mock_update.called_with(place)
```
KCGallagher, line 91: I can't find anything about this syntax online - is this meant to me `assert_called_with()`

KCGallagher (changes requested):
A few things but all really small so hopefully should take a few minutes to go through it all

pyEpiabm/pyEpiabm/core/parameters.py
```diff
@@ -160,12 +160,36 @@ def __init__(self):
...
+                                 "age_group3_min_age": [21, 21, 21, 21],
+                                 "age_group3_max_age": [65, 65, 65, 65],
+                                 "age_group3_prop": [0.005292966, .006331515,
+                                                     .064144615, 0.742058813],
+                                 "nearest_places": [3, 3, 6, 0],
+                                 "mean_size": [230, 1010,  3300, 14.28],
+                                 "max_size": [0, 0, 0, 5927],
+                                 "size_power": [0, 0, 0, 1.34],
+                                 "size_offset": [0, 0, 0, 5.35],
+                                 "mean_group_size": [25,  25,  100, 10],
```
rcw5890, line 186:
```suggestion
                                 "mean_group_size": [25, 25, 100, 10],
```

pyEpiabm/pyEpiabm/core/parameters.py
```diff
@@ -160,12 +160,36 @@ def __init__(self):
...
+                                 "age_group2_max_age": [5, 18, 21, 21],
+                                 "age_group2_prop": [1,	0.959727307,
+                                                     0.629021307,
+                                                     .040272693000000026],
+                                 "age_group3_min_age": [21, 21, 21, 21],
+                                 "age_group3_max_age": [65, 65, 65, 65],
+                                 "age_group3_prop": [0.005292966, .006331515,
+                                                     .064144615, 0.742058813],
+                                 "nearest_places": [3, 3, 6, 0],
+                                 "mean_size": [230, 1010,  3300, 14.28],
```
rcw5890, line 182:
```suggestion
                                 "mean_size": [230, 1010, 3300, 14.28],
```

pyEpiabm/pyEpiabm/property/place_foi.py
```diff
@@ -2,15 +2,16 @@
 # Calculate place force of infection based on Covidsim code
 #
 
-import pyEpiabm.core
 
 from .personal_foi import PersonalInfection
+from pyEpiabm.core import Parameters
```
KCGallagher, line 7:
```suggestion
from pyEpiabm.core import Parameters

from .personal_foi import PersonalInfection
```
Please maintain the PEP8 import order!

pyEpiabm/pyEpiabm/property/place_type.py
```diff
@@ -9,7 +9,11 @@ class PlaceType(Enum):
     """Enum representing a place"s type.
```
rcw5890, line 9:
```suggestion
    """Enum representing a place's type.
```

pyEpiabm/pyEpiabm/property/place_foi.py
```diff
@@ -35,9 +36,11 @@ def place_inf(place, infector, time: float):
             Infectiousness parameter of place
 
         """
+        transmission = PlaceInfection.params["place_transmission"]
+        place_ind = place.place_type.value - 1
+        num_groups = PlaceInfection.params["mean_group_size"][place_ind]
```
KCGallagher, line 41:
```suggestion
        place_idx = place.place_type.value - 1
        num_groups = PlaceInfection.params["mean_group_size"][place_idx]
```
`idx` is typically the standard abbreviation for index so this is slightly more readable

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,127 @@
...
+        time : float
+            Current simulation time
+
+        """
+
+        # Double loop over the whole population, clearing places
+        # of the variable population and refilling them.
+
+        # May want a param dict to be called in from file etc.
+        # place_params = Parameters.instance().place_params
```
KCGallagher, line 35: Do we still need this line of commented code?

rcw5890 (changes requested):

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,127 @@
...
+        add_temporary_population(0)
+
+    def create_age_weights(self, place, params):
+        """Function to return a list of people in the correct
+        age range to be added to a place, and the weights
+        for the different age groups.
+
+        Parameters
+        ----------
+
```
KCGallagher, line 90:
```suggestion
```

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,127 @@
...
+        place : Place
+            Place to add people to
+        params : dict
+            Dictionary of parameters with age structure data
+
+        Returns
+        -------
+
```
KCGallagher, line 98:
```suggestion
```

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,127 @@
...
+                    # Initialise the fixed population
+                    helper.update_place_group(place, group_size=ave_group_size,
+                                              person_list=person_list,
+                                              person_weights=weights,
+                                              mean_capacity=mean_cap)
+
+                elif place.place_type.name == "Workplace":  # WORKSPACE
+                    # Fixed population is initialised on first run
+                    # thinking of making sure people don't have more than one
+                    # workplace
```
KCGallagher, line 65: Is this handled in L119?

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,127 @@
...
+                   params["age_group3_max_age"][param_ind]]
+        prop = [params["age_group1_prop"][param_ind],
+                params["age_group2_prop"][param_ind],
+                params["age_group3_prop"][param_ind]]
+
+        person_list = []
+        weights = []
+        for person in place.cell.persons:
+            if (place.place_type in person.place_types):
+                # People can't have more than one place of each type.
```
KCGallagher, line 120: Open Q - if we reach here, is that an error that should be logged, or just normal functionality we don't need to worry about? Does it matter if the place doesn't get as many people as it was meant to?

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -0,0 +1,127 @@
...
+                continue
+            for i in range(3):
+                if (person.age > (min_age[i]-1) and person.age < max_age[i]):
+                    # Assumes age groups are distinct and integers.
+                    person_list.append(person)
+                    weights.append(prop[i])
```
KCGallagher, line 126: Is it worth breaking at the end of this, to make sure that people can't be added to multiple places (even if the age groups do overlap)? Or actually do we want this to be possible?

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -24,25 +28,104 @@ def __call__(self, time: float):
...
+
+        Parameters
+        ----------
+        place : Place
+            Place to change
+        mean_capacity : int
+            Average number people of in this place
+        power_law_params : list
+            Should only be given for workplaces, contains the further
+            parameters used to calculate place_size. List ent,ries should
```
KCGallagher, line 72:
```suggestion
            parameters used to calculate place_size. List entries should
```

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -24,25 +28,104 @@ def __call__(self, time: float):
...
+                           group_size: int = 0, person_list: list = None,
+                           person_weights: list = None):
+        """Specific method to update people in a place or place group.
+
+        Parameters
+        ----------
+        place : Place
+            Place to change
+        mean_capacity : int
+            Average number people of in this place
```
KCGallagher, line 69:
```suggestion
            Average number of people in this place
```

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -24,25 +28,104 @@ def __call__(self, time: float):
...
+
+        # Ensure that the number of people put in the place
+        # is at most its capacity or the total number of
+        # people in the cell. Will use a power law calculation if
+        # parameters are provided, and a Poisson distribution is not.
+        if power_law_params is None:
+            new_capacity = np.random.poisson(mean_capacity)
+        else:
+            assert len(power_law_params) == 3, \
+                "Incorrect number of power law parameters given"
```
KCGallagher, line 95:
```suggestion
                ("Incorrect number of power law parameters given"
                + " - should be of the form [maximum, offset, power]")
```

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -24,25 +28,104 @@ def __call__(self, time: float):
...
+
+        new_capacity = min(new_capacity, len(person_list))
+
+        if len(person_list) <= 0:
+            logging.warning("No people in the person list supplied.")
+        count = 0
+
+        try:
+            num_groups = np.random.poisson(math.ceil(new_capacity/group_size))
+        except ZeroDivisionError:
```
KCGallagher, line 110: Can you comment why this might occur? I assume it's when `group_size = 0` but not really sure why the function would be called with that, and why we'd want to handle it in this way

KCGallagher (changes requested):

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
Elizabeth-Hayman: nope

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
Elizabeth-Hayman: yes

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
Elizabeth-Hayman: I think normal functionality, because its looking at the whole cell each time, we wouldn't want to log as people are being perfectly normally put into workplaces etc and so being taken out of this list.

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
Elizabeth-Hayman: Think break is a good idea as stops people being added to the same place twice if they are boarderline on age

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
Elizabeth-Hayman: In the outdoor spaces or other places where there aren't groups then want trivially there to be one group in that place which is caught in the try/except

PR #88, opened 2022-03-07
Py editor config

### LINKED ISSUE #86: Editor config for consistent style

KCGallagher:
# Summary

Add `.editorconfig` file to specify style (and allow users to ensure that style on their local). Also update flake8 to give explicit ignores.

## Editor Config
* General requirements to strip whitespace from end of lines and add newline characters to end of files
* Python requirements to set indents as 4 spaces, as per PEP8 guidelines
* No C++ handling (see below)

See [here](https://www.freecodecamp.org/news/how-to-use-editorconfig-to-standardize-code-styles/) for more info on why it is a good idea. Note these are only conventions, and cannot be required in the CI pipeline.

## VS Code 
VS code (and almost every IDE in existence) support these files, and will use them in style warnings/autoformatting (depending on what your local setup has enabled). I recommend the EditorConfig [plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig).

## C++ Style
This has not been specified in the `.editorconfig` file - there are many options and I knew @NicholasFan235 would have opinions on what he preferred (or may not even want a consistent style across all files). If he does, Visual Studio has a long [list](https://docs.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019) of the possible options, and has recently added features to automatically detect this style from the code using [IntelliSense](https://visualstudiomagazine.com/articles/2021/02/11/vscode-cpp-update.aspx)*. I personally think it might be nice for consistency, but happy to defer to him for the final say there!

*This is a new feature, primarily developed in C# but might also support C++.

Resolves #86


I-Bouros (approved):

.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Python: PEP8 defines 4 spaces for indentation
+[*.py]
+indent_style = space
+indent_size = 4
+
+# Allow new paragrpah from two trailing whitespaces in markdown
```
rcw5890, line 16:
```suggestion
# Allow new paragraph from two trailing whitespaces in markdown
```

rcw5890, line 16: Actually, I don't even understand this comment. Is this the same as the trim_trailing_whitespace that is set below

rcw5890 (changes requested):
Looks very helpful thanks

.editorconfig
KCGallagher: So for most files we want to remove trailing whitespace at the end of lines, as it is meaningless. However, in markdown files, having two spaces at the end of a line generates a new paragraph, so it is meaningful. Therefore we want to get rid of trailing whitespace in all files, apart from markdown files, as it might actually be useful there.

rcw5890: Understood, thanks for the explanation

PR #92, opened 2022-03-08
Add transition time

lukedtc:
Added transition time class and update time to status change method in host progression. Tests also updated. Please ignore doc strings in wrong format for now will update imminently


pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -24,9 +23,10 @@ def __init__(self):
         transition matrix is set where each row of the matrix corresponds
         to a current infection status of a person. The columns of that
         row then indicate the transition probabilities for the remaining
```
KCGallagher, line 25:
```suggestion
        row then indicate the transition probabilities to the remaining
```

patricia-lamy:
I worked on the docstrings and they should be of the right format now.

I-Bouros (changes requested):

I-Bouros (commented):
Thank you @patricia-lamy for refactoring the documentation; I was going to make the same request before you merged the commit

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -119,18 +89,67 @@ def _update_next_infection_status(self, person: Person):
...
+        if (person.infection_status == InfectionStatus.Recovered or
+                person.infection_status == InfectionStatus.Dead):
+            transition_time = np.inf
+        else:
+            row_index = person.infection_status.name
+            column_index = person.next_infection_status.name
+            transition_time_icdf_object =\
+                self.transition_time_matrix.loc[row_index, column_index]
+            # Checks for susceptible to exposed case
+            #  where transition time is zero
```
KCGallagher, line 138: Is this the only case which could have a transition time of zero?

KCGallagher (changes requested):
Would it be possible to add a method to read in a given transition time matrix, or add a specific element to the matrix to make this more configurable? Given we have this in the transition matrix itself, it would be great to have that functionality here, although I see it might be hard to structure without an associated icdf (unless this is easy for a user to generate themselves?)

Otherwise this looks really great, I think we can get this merged in very soon!

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -119,18 +89,64 @@ def _update_next_infection_status(self, person: Person):
...
+            if transition_time_icdf_object == 0.0:
+                transition_time = 0.0
+            else:
+                transition_time =\
+                    transition_time_icdf_object.icdf_choose_noexp()
```
rcw5890, line 143:
```suggestion
            try:
                transition_time =\
                    transition_time_icdf_object.icdf_choose_noexp()
            except AttributeError as e:
                   if "object has no attribute 'icdf_choose_noexp'" in str(e):
                       transition_time = transition_time_icdf_object
                   else:
                       raise
```
This deserves more thought than I have given it but it may be a more graceful way to handle this.
It also has the advantage that if we want to run some simulation with transitions for which we want a single time, rather than a distribution, this allows us to just put those numbers in the matrix rather than having to contrive some delta function icdf.

rcw5890 (changes requested):
Looks nice thanks

pyEpiabm/pyEpiabm/tests/test_utility/test_transition_time_matrix.py
```diff
@@ -0,0 +1,41 @@
+import unittest
+import pandas as pd
+import numpy as np
+
+import pyEpiabm as pe
+from pyEpiabm.property.infection_status import InfectionStatus
```
KCGallagher, line 6:
```suggestion
from pyEpiabm.property import InfectionStatus
```

KCGallagher (commented):
@lukedtc I think the transition matrix should be moved to the init function in the sweep as we discussed - leaving it in core means that core is dependant on utilities which is undesirable as I think core should be as fundamental as possible (given utilities may have its own dependencies on core)

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
lukedtc: In CovidSim yes, this is the only instance with a transition time of zero.

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -112,13 +87,14 @@ def test_update_next_infection_status(self):
...
+                self.assertEqual(person.next_infection_status, None)
+            else:
+                self.assertEqual(person.infection_status,
+                                 person.next_infection_status)
 
-        # Check that method works for each infection status with all people
-        # going to dead infection status
         matrix = np.zeros([len(InfectionStatus), len(InfectionStatus)])
-        matrix[:, -1] = 1
+        matrix[:, -3] = 1
```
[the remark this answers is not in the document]
lukedtc: Implemented on your suggestion.

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -8,7 +8,9 @@
 import pyEpiabm as pe
 from pyEpiabm.core import Person
 from pyEpiabm.property import InfectionStatus
-from pyEpiabm.utility import InverseCdf, StateTransitionMatrix
+
+from pyEpiabm.utility import StateTransitionMatrix
+from pyEpiabm.utility import TransitionTimeMatrix
```
KCGallagher, line 13:
```suggestion
from pyEpiabm.utility import StateTransitionMatrix, TransitionTimeMatrix
```

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -126,18 +98,72 @@ def _update_next_infection_status(self, person: Person):
...
+            person.next_infection_status = None
+        else:
+            row_index = person.infection_status.name
+            weights = self.state_transition_matrix.loc[row_index].to_numpy()
+            outcomes = range(1, self.number_of_states + 1)
+
+            if len(weights) != len(outcomes):
+                raise AssertionError('The number of infection statuses must \
+                                    match the number of transition \
+                                    probabilities')
```
KCGallagher, line 112: Is this aligned right? And we generally avoid backslash as a line break where possible - string addition is preferred here

rcw5890 (changes requested):

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -126,18 +98,72 @@ def _update_next_infection_status(self, person: Person):
...
+        time : float
+            Current simulation time
 
-        next_infection_status_number = random.choices(outcomes, weights)[0]
-        next_infection_status = InfectionStatus(next_infection_status_number)
-        person.next_infection_status = next_infection_status
+        """
+        # Defines the transition time. If the person will not transition again,
+        # the transition time is set to infinity. Else, the transition time is
+        # defined using the TransitionTimeMatrix class, with a method choose
```
KCGallagher, line 137:
```suggestion
        # defined using the TransitionTimeMatrix class, with the method `choose()`
```

KCGallagher: (Check line limits on this)

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -126,18 +98,72 @@ def _update_next_infection_status(self, person: Person):
...
+            #  where transition time is zero
+            try:
+                transition_time =\
+                    transition_time_icdf_object.icdf_choose_noexp()
+            except AttributeError as e:
+                if "object has no attribute 'icdf_choose_noexp'" in str(e):
+                    transition_time = transition_time_icdf_object
+                else:
+                    print('a')
+                    raise
```
KCGallagher, line 157: Assume this is going to be deleted?

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -111,13 +84,17 @@ def test_update_next_infection_status(self):
...
+            else:
+                self.assertEqual(person.infection_status,
+                                 person.next_infection_status)
 
-        # Check that method works for each infection status with all people
-        # going to dead infection status
         matrix = np.zeros([len(InfectionStatus), len(InfectionStatus)])
-        matrix[:, -1] = 1
+        # Set ICU recovery infection status column values to 1. This way
+        # everyone who is not recovered or dead wiil have their next
```
KCGallagher, line 95:
```suggestion
        # everyone who is not recovered or dead will have their next
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -223,20 +239,25 @@ def test_call_specific(self):
...
-        # infection status multiple times in the same time step. This will be
-        # checked by setting the latent time as 0 so Person 1 should progress
-        # from susceptible to one if the infected statuses in the same time
-        # step.
-        pe.Parameters.instance().latent_period_iCDF = np.zeros(21)
+    @mock.patch('pyEpiabm.utility.InverseCdf.icdf_choose_noexp')
+    def test_multiple_transitions_in_one_time_step(self, mock_next_time):
+        """Reconfigure population and check that a person is able to progress
+        infection status multiple times in the same time step. This will be
+        checked by setting the time transition time as 0 so Person 1 should
```
KCGallagher, line 246:
```suggestion
        checked by setting the time transition time to 0 so Person 1 should
```

pyEpiabm/pyEpiabm/tests/test_utility/test_transition_time_matrix.py
```diff
@@ -0,0 +1,169 @@
...
+            row = TestInfectionStatus.Susceptiblesssss
+            column = TestInfectionStatus.Susceptiblesssss
+            test_mean = 1.0
+            test_icdf = np.ones(10)
+            matrix_object.update_transition_time_with_icdf(
+                row,
+                column,
+                transition_matrix,
+                test_icdf,
+                test_mean)
```
KCGallagher, line 114:
```suggestion
                row, column, transition_matrix,
                test_icdf, test_mean)
```
I don't like the really long function calls, fight me (or change the others too ;) )

pyEpiabm/pyEpiabm/utility/state_transition_matrix.py
```diff
@@ -70,7 +69,8 @@ def create_state_transition_matrix(self):
 
     def update_probability(self, current_infection_status_row: enum,
                            next_infection_status_column: enum,
-                           new_probability: float):
+                           new_probability: float,
+                           matrix: pd.DataFrame):
```
KCGallagher, line 73: I spy with my little eye, an argument that hasn't been documented.

pyEpiabm/pyEpiabm/utility/transition_time_matrix.py
```diff
@@ -0,0 +1,182 @@
...
+            InverseCdf(pe.Parameters.instance().mean_icurecov_to_recov,
+                       pe.Parameters.instance().icurecov_to_recov)
+        return matrix
+
+    def update_transition_time_with_float(
+                        self,
+                        current_infection_status_row: enum,
+                        next_infection_status_column: enum,
+                        matrix: pd.DataFrame,
+                        new_transition_time: float):
```
KCGallagher, line 91: okay please update the docstrings too

KCGallagher (changes requested):
Mostly issues with docstrings rather than code, which is definitely a good thing but probably won't make you any happier

rcw5890 (approved):

I-Bouros (approved):

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -126,18 +98,71 @@ def _update_next_infection_status(self, person: Person):
...
+            transition_time_icdf_object =\
+                self.transition_time_matrix.loc[row_index, column_index]
+            # Checks for susceptible to exposed case
+            # where transition time is zero
+            try:
+                transition_time =\
+                    transition_time_icdf_object.icdf_choose_noexp()
+            except AttributeError as e:
+                if "object has no attribute 'icdf_choose_noexp'" in str(e):
+                    transition_time = transition_time_icdf_object
```
KCGallagher, line 154:
```suggestion
                    transition_time = transition_time_icdf_object
                    assert isinstance(transition_time_icdf_object, (float, int)),
                        ("Entries of transition time matrix must either be ICDF" +
                        " objects or numbers")
```
If this doesn't cause flake errors I will be amazed - do check this works after

KCGallagher (changes requested):
Quite a few comments from last time were resolved but not addressed - its absolutely fine if you don't agree with suggests from the review but please leave a comment to explain why instead of just marking the comment as resolved without changing anything!

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -107,8 +79,8 @@ def set_infectiousness(person: Person):
         if person.infection_status == InfectionStatus.InfectASympt:
             infectiousness = init_infectiousness *\
                              pe.Parameters.instance().asympt_infectiousness
-        elif (person.infection_status == InfectionStatus.InfectMild or
-              person.infection_status == InfectionStatus.InfectGP):
+        elif (person.infection_status == InfectionStatus.InfectMild) or \
+             (person.infection_status == InfectionStatus.InfectGP):
```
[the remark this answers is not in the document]
lukedtc: Resolved now, change was made to appease flake8 but think it was messing up but reverted back to original.

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
lukedtc: Implemented

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -8,7 +8,8 @@
 import pyEpiabm as pe
 from pyEpiabm.core import Person
 from pyEpiabm.property import InfectionStatus
-from pyEpiabm.utility import InverseCdf, StateTransitionMatrix
+
```
KCGallagher, line 11:
```suggestion
```

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -23,64 +24,34 @@ def __init__(self):
...
-        assert latent_time >= 0.0, 'Negative latent time'
-
-        person.time_of_status_change = time + latent_time
+        # Instantiate parameters to be used in update transition time
+        # method
+        self.latent_to_symptom_delay =\
+            pe.Parameters.instance().latent_to_sympt_delay
+        self.model_time_step = 1 / pe.Parameters.instance().time_steps_per_day
+        self.delay = np.floor(self.latent_to_symptom_delay /
+                              self.model_time_step)
```
KCGallagher, line 54:
```suggestion
        self.delay = np.floor(self.latent_to_symptom_delay,
                              self.model_time_step)
```

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -126,18 +97,76 @@ def _update_next_infection_status(self, person: Person):
...
+                transition_time =\
+                    transition_time_icdf_object.icdf_choose_noexp()
+            except AttributeError as e:
+                if "object has no attribute 'icdf_choose_noexp'" in str(e):
+                    transition_time = transition_time_icdf_object
+                    assert isinstance(
+                        transition_time_icdf_object,
+                        (float, int)), \
+                        ("Entries of transition time matrix \
+                        must either be ICDF" + " objects or numbers")
```
KCGallagher, line 158:
```suggestion
                        (float, int)), \
                        ("Entries of transition time matrix"
                         + " must either be ICDF objects or numbers")
```
[PEP8](https://peps.python.org/pep-0008/#maximum-line-length) is really quite against backslashes so we should avoid them if possible, and they're definitely not needed here as you can break the comment across lines.

pyEpiabm/pyEpiabm/tests/test_utility/test_transition_time_matrix.py
```diff
@@ -0,0 +1,166 @@
...
+            row = None
+            column = None
+            test_mean = 1.0
+            test_icdf = np.ones(10)
+            matrix_object.update_transition_time_with_icdf(
+                row,
+                column,
+                transition_matrix,
+                test_icdf,
+                test_mean)
```
KCGallagher, line 123:
```suggestion
                row, column, transition_matrix,
                test_icdf, test_mean)
```

KCGallagher: Ditto for the three other five-line function calls

pyEpiabm/pyEpiabm/utility/state_transition_matrix.py
```diff
@@ -84,6 +84,9 @@ def update_probability(self, current_infection_status_row: enum,
             the column where the probability will be updated
         new_probability : float
             Updated transition probability value
+        matrix : pd.Dataframe
+            Input state transition matrix that will have
+            one of it's probabilities changed
```
KCGallagher, line 89:
```suggestion
            one of its probabilities changed
```

pyEpiabm/pyEpiabm/utility/transition_time_matrix.py
```diff
@@ -0,0 +1,188 @@
...
+            InverseCdf(pe.Parameters.instance().mean_icu_to_death,
+                       pe.Parameters.instance().icu_to_death_icdf)
+        matrix.loc['InfectICURecov', 'Recovered'] =\
+            InverseCdf(pe.Parameters.instance().mean_icurecov_to_recov,
+                       pe.Parameters.instance().icurecov_to_recov)
+        return matrix
+
+    def update_transition_time_with_float(
+                        self,
+                        current_infection_status_row: enum,
```
KCGallagher, line 88: Is this genuinely an enum (I think it should be `InfectionStatus`?)

pyEpiabm/pyEpiabm/utility/transition_time_matrix.py
```diff
@@ -0,0 +1,188 @@
...
+            entries changed
+        new_transition_time : float
+            Updated transition time value
+
+        """
+        try:
+            if (current_infection_status_row not in InfectionStatus) or \
+                    (next_infection_status_column not in InfectionStatus):
+                raise ValueError('Row and column inputs must be contained in\
+                                the InfectionStatus enum')
```
KCGallagher, line 114: Split strings over multiple lines using string addition (ditto for other value errors here)

KCGallagher (changes requested):
[PEP8](https://peps.python.org/pep-0008/#maximum-line-length) is really quite against backslashes - they should only be used if you don't have another option as they aren't handled well in errors or docs. 

In our work, the main valid use is when you are calling a long method, imported from another subpackage, so you can't fit it all on one line. Something like:

```python
parameter_value = pyEpiabm.Parameters.instance().\
    basic_reproduction_num
```
It can also be useful in long `assert` or `with` statements (although even this is being phased out).

It shouldn't be used in the following contexts:
* Function calls (Split the line after the comma)
* After a comma/ within brackets in general - (ditto)
* In complicated sums (use brackets and split before the operator)
* In complicated binary expressions (ditto with binary operator)
* With a long string (Split the string across lines with string addition)

Hope this helps - sorry to be a pain about it but it is really considered rather bad form now and some linters will be much stricter about them. Hopefully this helps for the future

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
lukedtc: This is a divide, not a backslash. Have changed to make line break before mathematical operator per PEP8

KCGallagher: Ahh that makes sense sorry - thanks for that!

KCGallagher (approved):

PR #96, opened 2022-03-09
Update docs

Elizabeth-Hayman:
Minor changes to the docs, but think I branched off recent work so might be bigger than expected


KCGallagher:
As this does contain lots of other work, lets wait for that to be merged before we merge this one

KCGallagher:
@Elizabeth-Hayman Have verified there is nothing added from this PR, so closing it.

PR #99, opened 2022-03-09
Py infection forces

KCGallagher:
Includes household infection force only 
Also restructures repo to have FOI calculation in property subpackage


pyEpiabm/pyEpiabm/property/personal_foi.py
```diff
@@ -0,0 +1,54 @@
+#
+# Calculate infectiousness and susceptibility for an individual
+#
+
+
+class PersonalInfection:
+    """Class to calculate the infectiousness and susceptibility
+    parameters an individual.
```
rcw5890, line 8:
```suggestion
    parameters of a Person.
```

rcw5890, line 8:
```suggestion
    parameters of a :class:`Person`.
```
This might work as well, and should create a hyperlink in the docs

rcw5890 (changes requested):

rcw5890 (approved):

PR #106, opened 2022-03-11
Infectiousness progression

patricia-lamy:
Add infectiousness progression to the model. An infected person gets assigns an initial infectiousness, which is updated every time step while they are infectious, following the infectiousness profile that is fed in the model.


pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -66,22 +109,30 @@ def set_infectiousness(person: Person):
...
+        time : float
+            Current simulation time
 
         Returns
         -------
         float
-            Infectiousness of a person
+            Initial infectiousness of a person
+        float
+            Infection start time of a person
```
KCGallagher, line 120: Does it return these? I can't see it in the function call

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -190,6 +281,8 @@ def __call__(self, time: float):
                             [InfectionStatus.InfectASympt,
                              InfectionStatus.InfectMild,
                              InfectionStatus.InfectGP]:
-                        self.set_infectiousness(person)
+                        self.set_infectiousness(person, time)
                     self._update_next_infection_status(person)
                     self._update_time_status_change(person, time)
+                    continue
```
KCGallagher, line 287: I think this is being misused - continue skips to the next iteration of the top level for/while loop. Here it is at the end of a while loop, so won't do anything. Do you also want to skip on to the next person?

pyEpiabm/pyEpiabm/sweep/initial_infected_sweep.py
```diff
@@ -33,6 +33,11 @@ def __call__(self, sim_params: dict):
             raise ValueError('Initial number of infected people needs to be \
                                             less than the total population')
 
+        start_time = sim_params["simulation_start_time"]
+        if start_time < 0:
+            raise ValueError('Simulation start time needs to be greater or\
+                              equal to 0')
```
KCGallagher, line 39:
```suggestion
            raise ValueError('Simulation start time needs to be greater or'
                             + 'equal to 0')
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,85 @@ def test_icdf_exception_raise(self):
         test_sweep.transition_time_matrix.loc[row_index, column_index].\
             icdf_choose_noexp.assert_called_once()
 
+    def test_infectiousness_progression(self):
+        """Tests that the output is a numpy ndarray and that the tail of the
+        array is 0, starting at the element k.
+        """
+        # Parameters to determine where the tail starts
+        infectious_period = pe.Parameters.instance().asympt_infect_period
+        model_time_step = 1 / pe.Parameters.instance().time_steps_per_day
```
KCGallagher, line 198: If these scalars are just parameters, can we put the values in here rather than having to rely on their values in the parameter file. Will be especially pertinent once we are reading in parameters from an external file we don't want to be involved in the tests

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,85 @@ def test_icdf_exception_raise(self):
...
+        self.assertTrue((tail == zeros).all())
+
+    def test_infectiousness_progression_small_time_steps(self):
+        """Tests that an assertion error is raised if the model time steps
+        length is too small (ie the time steps per day is too big).
+        """
+        # Stocks the real time steps per day value
+        real = pe.Parameters.instance().time_steps_per_day
+
+        # Assigns temporarily a new value for time steps per day to raise error
```
KCGallagher, line 217: Might be better to have a mock here - then you won't have changed the real value in case this test bombs out

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -52,11 +56,50 @@ def __init__(self):
         self.delay = np.floor(self.latent_to_symptom_delay
                               / self.model_time_step)
 
+        # Infectiousness progression
+        # Instantiate parameters to be used in update infectiousness
+        infectious_profile = pe.Parameters.instance().infectiousness_prof
+        inf_prof_res = len(infectious_profile) - 1
```
KCGallagher, line 62: I'm confused what this is? If this is just an index for the last element in the array, it is more pythonic to use the index -1.

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,85 @@ def test_icdf_exception_raise(self):
...
+        # Parameters to determine where the tail starts
+        infectious_period = pe.Parameters.instance().asympt_infect_period
+        model_time_step = 1 / pe.Parameters.instance().time_steps_per_day
+        k = int(np.ceil(infectious_period / model_time_step))
+        # Initialisation
+        test_sweep = pe.sweep.HostProgressionSweep()
+        infect_prog = test_sweep.infectiousness_progression
+        # Checks output type is numpy array
+        self.assertIsInstance(infect_prog, np.ndarray)
+        # Checks elements are 0 after k
```
KCGallagher, line 205: Should we check values before this are all non-zero? Given you're making an array of zeros initially, I think this would be a more useful test

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -52,11 +56,50 @@ def __init__(self):
...
+        infectious_period = pe.Parameters.instance().asympt_infect_period
+        # Extreme case where model time step would be too small
+        max_inf_steps = 2550
+        # k is number of time steps a person is infectious
```
KCGallagher, line 66: Could we use a more meaningful name than k in that case - covidsim likes the one letter variables but I think we are trying to avoid them here!
(Ditto for j, s etc)

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,85 @@ def test_icdf_exception_raise(self):
...
+        pe.Parameters.instance().time_steps_per_day = real
+
+    def test_limit_case_infectiousness_progression(self):
+        """Tests the case in the infectiousness progression method where the
+        parameter j would be greater or equal to the infectiousness profile
+        resolution. In that case, the elements of the array infectiousness
+        progression are simply set to 0.
+        """
+
+        with mock.patch('numpy.floor') as mock_floor:
```
KCGallagher, line 232: Good practice to check a mock is being called the expected number of times - I think it this case it might pick up that the HostProgressionSweep also calls it once (which should be avoided if it isn't meant to be mocked here) which may mean you have to create the mock later

KCGallagher (changes requested):
This looks really great - thank you so much for tackling this! No major suggestions, but some smaller (non-structural) things. I think it would be great to have an explicit test for multiple time steps per day, as I think that this is where the sweep could go wrong most easily (skipping forwards too many infectiousness values n a set number of days)

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
patricia-lamy: No indeed, I was confused with the documentation formatting.

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
patricia-lamy: I removed it

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: I think we would want these to be read from the parameter file (and from an external file if it becomes our way to input parameters), as these are the kind of parameters that users might want to change easily.  The model time step really need to stay linked to time steps per day, which is used elsewhere in the code.

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: Yeah that's what I tried to do at first but I didn't find how to mock a Parameter instance. This was the solution.

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
patricia-lamy: It's the infectiousness profile resolution, it's a parameter equal to the length minus 1 of the infectious_profile array, used elsewhere than just for the infectious_profile array.

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: Yup!

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
patricia-lamy: alright

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: It's currently called 15 times. Now that it the thing I want to mock i in the __init__ I don't know how I can call it later ?? :(

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: I think I need to call them from param, because these same parameters are called from the parameter file in the init of the host-progression-sweep, and the test would fail if the user changes the params.

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: I tried to do a try exception block!

KCGallagher:
We had a discussion about mocking an instance on the `Parameters` class - I know you avoided that in this PR but I wondered if you have tried this method - worked fine for me in my person tests:

```python
with patch('pyEpiabm.Parameters.instance') as mock_param:
    mock_param.return_value.use_ages = False
    self.person.set_random_age()
    mock_param.assert_called_once()
    self.assertEqual(self.person.age, 40)
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
KCGallagher: See my comment for a way to mock Parameters instead!

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
KCGallagher: Yes absolutely in the main code, but best not to rely on an external module in the tests (particularly as it will soon import in external values) when you could simply copy the floats across here!

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,116 @@ def test_icdf_exception_raise(self):
...
+            mock_floor.return_value = 57
+            test_sweep = pe.sweep.HostProgressionSweep()
+            infect_prog = test_sweep.infectiousness_progression
+            # Checks that the output is a numpy array
+            self.assertIsInstance(infect_prog, np.ndarray)
+            # Checks that all elements are equal to 0
+            zeros = np.zeros(2550)
+            self.assertTrue((infect_prog == zeros).all())
+            # Checks that it is called the right number of times
+            self.assertEqual(mock_floor.call_count, num_infectious_ts + 1)
```
KCGallagher, line 273: Needs a comment to explain the +1 and why its okay (i.e. that it is also called in the init function, but for an attribute that isn't used here)/ Also worth checking with @rcw5890 if there's a better way to do this!

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,116 @@ def test_icdf_exception_raise(self):
...
+            person.initial_infectiousness = 1.1
+            person.infection_start_time = 0
+        # Updates infectiousness after one day
+        for person in self.people:
+            test_sweep._updates_infectiousness(person, 1)
+        # Assert that non infectious people have infectiousness of 0
+        self.assertEqual(self.people[0].infectiousness, 0)
+        self.assertEqual(self.people[1].infectiousness, 0)
+        self.assertEqual(self.people[8].infectiousness, 0)
+        self.assertEqual(self.people[9].infectiousness, 0)
```
KCGallagher, line 296:
```suggestion
        for i in [0, 1, 8, 9];
            self.assertEqual(self.people[i].infectiousness, 0)
```

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,116 @@ def test_icdf_exception_raise(self):
...
+            infect_prog = test_sweep.infectiousness_progression
+            # Checks that the output is a numpy array
+            self.assertIsInstance(infect_prog, np.ndarray)
+            # Checks that all elements are equal to 0
+            zeros = np.zeros(2550)
+            self.assertTrue((infect_prog == zeros).all())
+            # Checks that it is called the right number of times
+            self.assertEqual(mock_floor.call_count, num_infectious_ts + 1)
+
+    def test_updates_infectiousness(self):
```
KCGallagher, line 275: Can we have an explicit test for when there are multiple timesteps per day, to make sure people don't progress too fast (i.e. they still step on once per day not once per time step). And make clear if you also support multiple days per timestep (I think we do so far in the rest of the codebase)

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -29,6 +29,10 @@ def __init__(self):
         is also initialised and associated parameters are called from the
         parameters class.
 
+        Infectiousness progression defines an array to scale a person's
+        infectiousness that depends on time since the start of the infection,
+        measured in timesteps (following what is done in Covidsim).
```
I-Bouros, line 34:
```suggestion
        Infectiousness progression defines an array used to scale a person's
        infectiousness and which depends on time since the start of the
        infection, measured in timesteps (following what is done in Covidsim).
```

I-Bouros (approved):

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: I still think we need it, because if we change the parameters later on, then the test would fail because these two parameters would not match those that are taken in the main code.

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: See test_infectiousness_progression_small_time_steps

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: Now that I can easily mock the parameters instance everything is mocked in the test, so that should solve this.

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -48,15 +52,60 @@ def __init__(self):
...
+        infectious_profile = pe.Parameters.instance().infectiousness_prof
+        inf_prof_resolution = len(infectious_profile) - 1
+        infectious_period = pe.Parameters.instance().asympt_infect_period
+        # Extreme case where model time step would be too small
+        max_inf_steps = 2550
+        # Define number of time steps a person is infectious:
+        num_infectious_ts =\
+            int(np.ceil(infectious_period / self.model_time_step))
+        if num_infectious_ts >= max_inf_steps:
+            raise ValueError('Number of infect timesteps exceeds limit')
```
KCGallagher, line 72: I'm not sure what you mean here but I don't think its clear - can you rephrase the error message? Maybe "Number of timesteps in infectious period exceeds limit - please reduce `time_steps_per_day`"?

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -48,15 +52,60 @@ def __init__(self):
...
+                    (infectious_profile[associated_inf_value] * (1 - t)
+                     + infectious_profile[associated_inf_value + 1] * t)
+                sum_for_scaling += infectiousness_prog[i]
+            else:  # limit case where we define infectiousness to 0
+                infectiousness_prog[i] =\
+                    infectious_profile[inf_prof_resolution]
+                sum_for_scaling += infectiousness_prog[i]
+        # Scaling
+        scaling_param = sum_for_scaling / num_infectious_ts
+        # Removes artificial case of division by 0, would only happen if every
```
KCGallagher, line 95:
```suggestion
        # Avoids artificial case of division by 0, would only happen if every
```
I think this better describes what you are doing?

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -66,22 +115,23 @@ def set_infectiousness(person: Person):
...
+        time : float
+            Current simulation time
 
         """
         init_infectiousness = np.random.gamma(1, 1)
         if person.infection_status == InfectionStatus.InfectASympt:
-            infectiousness = (init_infectiousness
-                              * pe.Parameters.instance().asympt_infectiousness)
+            infectiousness = init_infectiousness *\
+                             pe.Parameters.instance().asympt_infectiousness
```
KCGallagher, line 125:
```suggestion
            infectiousness = (init_infectiousness *
                              pe.Parameters.instance().asympt_infectiousness)
```
See [this](https://github.com/SABS-R3-Epidemiology/epiabm/pull/92#pullrequestreview-905912641) comment to Luke about backslashes - N.B this needs changing in multiple places

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -66,22 +115,23 @@ def set_infectiousness(person: Person):
...
               person.infection_status == InfectionStatus.InfectGP):
-            infectiousness = (init_infectiousness
-                              * pe.Parameters.instance().sympt_infectiousness)
-        person.infectiousness = infectiousness
+            infectiousness = init_infectiousness *\
+                             pe.Parameters.instance().sympt_infectiousness
+            person.initial_infectiousness = infectiousness
+        person.infection_start_time = time
+        if person.infection_start_time < 0:
+            raise AssertionError('The infection start time cannot be negative')
```
KCGallagher, line 134: This is  'ValueError` really as you're not calling the assert explicitly

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -176,6 +189,138 @@ def test_icdf_exception_raise(self):
...
+            mock_param.return_value.asympt_infect_period = 14
+            infectious_period = pe.Parameters.instance().asympt_infect_period
+            model_time_step = 1 / pe.Parameters.instance().time_steps_per_day
+            num_infectious_ts = int(np.ceil(infectious_period /
+                                            model_time_step))
+            test_sweep = pe.sweep.HostProgressionSweep()
+            infect_prog = test_sweep.infectiousness_progression
+            # Checks output type is numpy array
+            self.assertIsInstance(infect_prog, np.ndarray)
+            # Checks elements are 0 after k and greater than 0 before
```
KCGallagher, line 238: This new test is really great, but maybe update this comment now we have renamed `k`?

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
KCGallagher: i might be misunderstanding, but that test only checks that the array is filled in correctly doesn't it? It would be good to check with the host sweep itself that people there only change infectiousness on full days - its possible otherwise a person could step through the array at the wrong rate (I know this implementation works but I think its something that could go wrong very easily if we ever have to refactor here for any reason)

KCGallagher (changes requested):
Looks good, few small things!

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
patricia-lamy: I removed those I inserted that could be done without breaking flake 8.

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
patricia-lamy: I'm not sure I understand, why would we want the infectiousness to only be updated on full days if we have smaller time steps? Maybe there is a functionality that I don't understand in the code, this is not what I understood from my reading of CovidSim. 

Yes the test only checks that the infectiousness progression array is filled in correctly (this was the goal of the test). Are you suggesting that in test_updates_infectiousness there should be a subtest with the case where the time step is smaller than 1?

pyEpiabm/pyEpiabm/tests/test_sweep/test_host_progression_sweep.py
```diff
@@ -185,12 +196,268 @@ def test_icdf_exception_raise(self):
...
+    def test_infectiousness_progression(self):
+        """Tests that the output is a numpy ndarray and that the tail of the
+        array is 0, starting at the first element after the last infectious
+        time step, when the number of time steps per day is 1 and the
+        infectious period is 14 days.
+        """
+        with mock.patch('pyEpiabm.Parameters.instance') as mock_param:
+            mock_param.return_value.time_steps_per_day = 1
+            mock_param.return_value.asympt_infect_period = 14
+            mock_param.return_value.infectiousness_prof =\
```
KCGallagher, line 214: Should be able to read these in from file now

KCGallagher (changes requested):

I-Bouros (approved):

KCGallagher (approved):

rcw5890 (changes requested):

pyEpiabm/pyEpiabm/sweep/host_progression_sweep.py
```diff
@@ -52,15 +56,56 @@ def __init__(self):
...
+                              self.model_time_step)
+
+        # Infectiousness progression
+        # Instantiate parameters to be used in update infectiousness
+        infectious_profile = pe.Parameters.instance().infectiousness_prof
+        inf_prof_resolution = len(infectious_profile) - 1
+        inf_prof_average = np.average(infectious_profile)
+        infectious_period = pe.Parameters.instance().asympt_infect_period
+        # Extreme case where model time step would be too small
+        max_inf_steps = 2550
```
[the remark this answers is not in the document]
patricia-lamy: I removed this from the param file, as well as inf_prof_res that was in the param file and that was not used anymore.

rcw5890 (approved):

PR #118, opened 2022-03-16
Py profiler

### LINKED ISSUE #114: pyEpiabm - profiling
Investigate profiling on pyEpiabm backend
Produce summary of expensive functions
Consider permentant inclusion of profiling capability in module (if dependancies allow - could be in devs only)

KCGallagher:
Instructions for profiler, and relevant files in `.gitignore`

Resolves #114


CONTRIBUTING.md
```diff
@@ -110,5 +110,31 @@ We have some general conventions in our `.editorconfig` [file](https://github.co
...
+python3 kernprof.py -l python_examples/simulation_flow.py
+```
+
+This will create an output file in the current working directory with the same name as the file you have profiled, but an added `.lprof` extension. To view this, use:
+
+```console
+python3 -m line_profiler simulation_flow.py.lprof 
+```
+
+In this way you can indentify bottlenecks in performance, or compare the time spent running your functions to the overall simulation time. Note these is a small performance cost in profiling (and the `@profile` decorator will throw errors for users who don't have this module installed), so all references to profiling in the code should be removed before submission.
```
rcw5890, line 137:
```suggestion
In this way you can identify bottlenecks in performance, or compare the time spent running your functions to the overall simulation time. Note there is a small performance cost in profiling (and the `@profile` decorator will throw errors for users who don't have this module installed), so all references to profiling in the code should be removed before submission.
```

rcw5890 (changes requested):

rcw5890 (approved):

PR #132, opened 2022-04-28
Cpp spatial weights

### LINKED ISSUE #65: cEpiabm - Spatial weightings
Replicate python code in C++ - a lot of the framework should be there already! P.S make sure to branch off Nic's PR #36 to have the up to date version

Elizabeth-Hayman:
Spatial sweep with distance weights
 Closes #65


KCGallagher:
@Elizabeth-Hayman is this still a draft or are you wanting a review? It looks like this has branched off a more recent version of cEpiabm than the version on main, so may want to wait until that is merged in before looking at this. Also there are a lot of conflicts so probably worth resolving those before any reviews

cEpiabm/src/sweeps/household_sweep.cpp
```diff
@@ -81,7 +81,7 @@ namespace epiabm
 
         double infectiousness = Covidsim::CalcHouseInf(infector, timestep);
         double susceptibility = Covidsim::CalcHouseSusc(infector, infectee, timestep);
-        double foi = infectiousness * susceptibility;
+        double foi = infectiousness * susceptibility * 0.5;
```
KCGallagher, line 84: Think you have accidentally merged in an old version of the C++ - @NicholasFan235 changed this the other way recently? Maybe worth checking no other changes have been reversed?

cEpiabm/src/sweeps/spatial_sweep.cpp
```diff
@@ -16,30 +18,68 @@ namespace epiabm
...
+        if (doDistance)
+        {
+            std::pair<double, double> current_loc = currentCell->location();
+            for (Cell &cell : cells)
+            {
+                weightVector.push_back(1./DistanceMetrics::Dist(cell.location(), current_loc));
+            }
+        }
+        // LCOV_EXCL_START
+        else if (doCovidsim) //this will be in the config file in cpp-dev
```
KCGallagher, line 45: Query - is `cpp-dev` a separate branch? Does that mean we need that branch in for this to work?

cEpiabm/src/utility/distance_metrics.cpp
```diff
@@ -0,0 +1,28 @@
...
+    }
+            
+    double DistanceMetrics::DistEuclid(location loc1, location loc2)
+    {
+        return pow(pow(loc1.first-loc2.first, 2) + pow(loc1.second-loc2.second, 2), 0.5);
+    }
+            
+    double DistanceMetrics::DistCovidsim(location /*loc1*/, location /*loc2*/)
+    {
+        return 2;
```
KCGallagher, line 24: Is this a placeholder?

cEpiabm/test/utility/test_distance_metrics.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+TEST_CASE("utility/distance_metrics: test distance metric", "[DistanceMetrics]")
+{   // tests the Dist function returns the same as the DistEuclid
+    auto loc = std::make_pair(1.0, 1.0);
+    double subject = DistanceMetrics::Dist(loc, loc);
+    double euclid = DistanceMetrics::DistEuclid(loc, loc);
+    REQUIRE(subject == euclid);
+
+    auto loc2 = std::make_pair(1.0, 0.0);
```
KCGallagher, line 19:
```suggestion
    auto loc2 = std::make_pair(-1.0, 2.0);
```
Tiny thing but if we have two distances it might be good for them to have x and y separation to check this is initialised properly (and crossing the origin for the same reason). I know we aren't testing to the same level on the C++ but thought I'd suggest this as it seems an easy change to make! Ditto on the test cases below

KCGallagher (changes requested):
I wasn't sure if this was ready for review, but had some time so thought I would have a look - please ignore comments here if it is still WIP though! 
Overall it looks really great - I added a few minor comments/questions but nothing that would take more than 5 minutes to go through!

One thing that is slightly concerning - @NicholasFan235 this seems to reverse your change on the factor of 0.5 in the FOI? I have flagged it, and can't see anything else, but maybe worth going through to check no other recent changes have been overwritten

cEpiabm/src/utility/distance_metrics.cpp
Elizabeth-Hayman: Very much so

cEpiabm/src/sweeps/spatial_sweep.cpp
Elizabeth-Hayman: doCovidsim is a true/false flag that for a simulation needs to be initialised in a config file somewhere. But everything will work with the tests just hardcoding the flag.

cEpiabm/src/sweeps/household_sweep.cpp
Elizabeth-Hayman: so we DON'T want the factor of 0.5?

KCGallagher: No, but I think it must have been added by merging back in an old branch so worth checking with @NicholasFan235  whether any other changes have been overwritten too

cEpiabm/src/sweeps/spatial_sweep.cpp
KCGallagher: Oh that makes sense - would the config file add it to a global namespace or would it need to be read in from parameters here?

cEpiabm/src/utility/distance_metrics.cpp
KCGallagher: How easy would it be to add the covidsim implementation, so we can compare to them directly?

cEpiabm/src/sweeps/spatial_sweep.cpp
NicholasFan235: Fixed this in the cpp-dev branch, happy for this to go into main first and then we can pull in cpp-dev

cEpiabm/src/utility/distance_metrics.cpp
KCGallagher: Okay we can get this in but given this is an unmarked placeholder I will also make an issue and allocate it to you so we don't forget!

KCGallagher (approved):

rccreswell (approved):

PR #133, opened 2022-06-22
Latest developmetn branch ready to merge to main

NicholasFan235:
Host progression sweep
Toy population factory class
Infection parameters added including inverse cdf and other distributions (age based)
Unified random number handling with multithreading support
Configuration classes with json config reader
Ability to run multithreaded simulation


cEpiabm/src/simulations/basic_simulation.hpp
```diff
@@ -42,4 +42,4 @@ namespace epiabm
 }
 
 
-#endif // EPIABM_SIMULATIONS_BASIC_SIMULATION_HPP
+#endif // EPIABM_SIMULATINS_BASIC_SIMULATION_HPP
```
KCGallagher, line 45: If this change intentional? I know it is spelt 'SIMULATINS' in a few places now but just wanted to check!

KCGallagher (approved):
This is incredible work @NicholasFan235, thanks so much for all your effort on this! I think it looks great, happy to merge this as soon as #132 is in.

Github still gives warnings about missing newline characters at the end of files, but as I think all modern compilers handle this automatically I don't think it needs changing.

PR #134, opened 2022-06-23
Py outputter

patricia-lamy:
This is not ready to merge. Please read this piece here before starting reviewing the code.

Here is Lizzi's work and my work on adding age_group in the counting of infections. The main modification is, instead of working with an array for the number of people in each infection status, we work with an array of arrays (in the array with each infection status, each element is an array with the number of people for each age_group).

In the case where age is not considered, we still have an array of arrays, with only one age group (everyone is in the same age_group). This structure was chosen so we have a consistent format for the outputs. 

This modification impacts bits of code a bit everywhere. I think we covered most of it, but it would be nice to have additional pairs of eyes to go through. 

The main modifications are in core/_compartment_counter.py, and with an important bit in the routine/simulation.py. For the writing of the csv output, we add an extra column if age is considered in the model, with many extra rows (each containing the count for each infection status, same structure as what was chosen when spatiality was added). If age is not considered, the extra column will not be written into the csv.

Now, I think we covered most changes in the code, but there is still a lot of testing to be done. Here is what needs to be done:

- I have two tests that are failing in test_simulation.py : test_write_to_file and test_s_write_to_file. That is because of the mock.assert_called_with. Any help would be appreciated.
- There needs to be additional testing on a lot of the things we modified. Even if, for some, the coverage will be 100%, I don't think it covers certain cases of code that we added. So a thorough look at what needs to be added in the tests would be needed.
- I have not looked into if this actually works and outputs the right thing, this would be a nice thing to do too. 

These notes are for future Patricia and Lizzi, but also for anyone who is looking to work on the code to make some progress into what we need to do before the end of the summer. So please feel free to start diving into this branch, but please write on here (github or teams chat) what you are doing so that we can all know what is going on. I think I summed up and explained how you can help.


KCGallagher:
"_I have two tests that are failing in test_simulation.py : test_write_to_file and test_s_write_to_file. That is because of the mock.assert_called_with. Any help would be appreciated._"

Am I missing something here - there's no issues with the mock calls in the unit test log? You have one error (from an assertion call you haven't handled correctly) and three failures - two are runtime errors from not setting a config file for the parameters and one is a key error because you haven't defined age groups in your updated compartment counter testing. With all these issues (particularly the config file) there are lots of examples in the tests of the correct method (you should inherit from my custom test class) so I'd recommend checking those out - the `test_simulation.py` file is probably a good place to start!

KCGallagher (commented):
Following on from yesterday's meeting, I thought I'd have a quick look through this PR to head off any issues early.
I'm still not certain I understand the benefits of this approach over the more modular approach taken in the C++ (whereby the choice of outputter does not affect the internal code structure, and you can simply bolt on additional outputer classes without having to rewrite the rest of the code). I am concerned this adds unnecessary complexity, and think we would certainly need to test this extensively to ensure other functionality is not affected. I'm also concerned this could have a significant negative affect on performance - if you continue with this approach can you benchmark CPU and memory usage against the old configuration?

You may find it is actually simpler to bolt a single outputter function onto the antecedent code, which can iterate through all individual of a population and record their infection status in an age-stratified table without relying on the compartment counter. This has the significant advantage of only storing this information when it is required (this reporter would only be called every T timesteps, where T will likely be large). I am not worried about the additional CPU cost of this O(N) process (in contrast to the majority of our sweeps), while I think the reduction in memory cost could be significant. Let me know what you think of this, or if there are downsides to this approach that I have missed!

patricia-lamy:
We have working simulations with age stratified output, see simulation_flow_withage.py in the folder python_example. This simulation, when summing over the age groups, yields the same results as the simulation without age stratification. 

We also have a working simulation with age and spatial stratification (but plotting only the sum of all group ages). The spatial simulation already in place doesn't work right now, because of the input population file (is not compatible yet with the input needed for age stratification). 

Next steps: make additional verification that the outputs we obtain are comparable to outputs without age stratification. Look at speed of computation. Add functional tests. Fix spatial simulation flow in this environment. Get up the coverage of tests to 100%.

pyEpiabm/pyEpiabm/tests/test_unit/test_utility/test_distance_metrics.py
```diff
@@ -3,12 +3,18 @@
...
 
-class TestDistanceFunctions(unittest.TestCase):
+class TestDistanceFunctions(TestMockedLogs):
     """Test the 'DistanceFunctions' class.
     """
 
+    @classmethod
+    def setUpClass(cls) -> None:
+        super(TestDistanceFunctions, cls).setUpClass()
+        # Sets up patch on logging
```
KCGallagher, line 16: Standard configuration so doesn't need commenting (ditto elsewhere)

python_examples/age_simulation.py
```diff
@@ -0,0 +1,93 @@
+#
+# Example simulation script with data output and visualisation
+#
+
+import os
+import logging
+from age_stratified_plot import Plotter
```
KCGallagher, line 7: Follow PEP8 import order guidelines (previously linked)

python_examples/age_simulation.py
```diff
@@ -0,0 +1,93 @@
...
+logging.basicConfig(filename='sim.log', filemode='w+', level=logging.DEBUG,
+                    format=('%(asctime)s - %(name)s'
+                            + '- %(levelname)s - %(message)s'))
+
+# Set config file for Parameters
+dirname = os.path.dirname(os.path.abspath(__file__))
+paramfile = (dirname.split(os.sep))
+paramfile = paramfile[1:-1]
+paramfile.append("spatial_example/spatial_parameters.json")
+paramfile = os.path.join('C:\\', *paramfile)
```
KCGallagher, line 20: As previously discussed is this really needed? Please try the alternative I sent you @Elizabeth-Hayman - if it doesn't work we can debug it but this shouldn't be necessary

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -0,0 +1,176 @@
...
+import matplotlib.pyplot as plt
+import os
+
+# csv input files should have the column headers:
+# time, infection_Status1, infection_status2, ..., age_range
+# so there will be multiple entries for each timepoint.
+
+
+class Plotter():
+    """Funtion to take a csv file and return various plots,
```
KCGallagher, line 15: Docstring should describe the class not a function

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -0,0 +1,176 @@
...
+        # with the substring "age"
+        self.age_name = next((s for s in list(self.data.columns)
+                             if "age" in s.lower()), None)
+        self.do_ages = (self.age_name is not None)
+        if self.do_ages and (self.age_list is None):
+            num = self.data[self.age_name].max()
+            self.age_list = [str(5*i)+"-"+str(5*i+5) for i in range(num+1)]
+
+    def sum_infectious(self) -> None:
+        """Helper function which sums across all columns containing infectious
```
KCGallagher, line 51: Is this helper function intended to be used externally? Consider marking it as private if not

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -0,0 +1,176 @@
...
+    def dates(self, dataFrame=None, period='weekly') -> None:
+        """Helper function to calculate a dictionary associating each
+        timestep, with the date of a day, or week.
+        Parameters
+        ----------
+        dataFrame : DataFrame
+            dataFrame containing a "time" column
+        period : str
+            Either "daily" or "weekly"
+        Returns
```
KCGallagher, line 68:
```suggestion
            Either "daily" or "weekly"

        Returns
```

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -0,0 +1,176 @@
...
+            if period == 'daily':
+                date_list.append((start + pd.DateOffset(days=time)).date())
+            else:
+                if time % 7 == 0:
+                    week_date = (start + pd.DateOffset(days=time)).date()
+                date_list.append(week_date)
+        date_list = [d.strftime('%m-%d') for d in date_list]
+        return date_list
+
+    def fiveToTen(self):
```
KCGallagher, line 90: Function name does not obviously indicate what function does (i.e. Convert between/combine different age groups)

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -0,0 +1,176 @@
...
+        infection_category : str
+            Category to be plotted, defaults to Total Infectious
+        write_Df_toFile : str
+            Optional argument, .csv filepath if configured plotting
+            data should be written to file
+        """
+        if infection_category == "Total Infectious":
+            self.sum_infectious()
+        if self.age_list[0] == '0-5':
+            # If the first age range is '0-5'
```
KCGallagher, line 123: Comment is redundant

python_examples/age_stratified_example/gibraltar_parameters.json
```diff
@@ -0,0 +1,407 @@
+{
+    "basic_reproduction_num": 2.8,
```
KCGallagher, line 2: Is this a direct duplication of the other Gibraltar parameters file? If so, just read in this other file as there's no need for a separate file. 

(In general if you consider this as a sub example of the Gibraltar simulation with age stratification, which I think it is, I think it would be best to restructure the directory (moving these workflows into a subfolder within that directory) to reflect that)

python_examples/age_stratified_example/input.csv
```diff
@@ -0,0 +1,401 @@
+cell,microcell,location_x,location_y,household_number,Susceptible
```
KCGallagher, line 1: Ditto if you're using the same initial population as the Gibraltar example

python_examples/age_stratified_example/simulation_flow_age_spatial.py
```diff
@@ -0,0 +1,116 @@
...
+# Currently running with Kit's gibraltar set-up
+#
+
+import os
+import logging
+import pandas as pd
+import matplotlib.pyplot as plt
+
+import pyEpiabm as pe
+from age_stratified_plot import Plotter
```
KCGallagher, line 12:
```suggestion
import pyEpiabm as pe

from age_stratified_plot import Plotter
```

python_examples/age_stratified_example/simulation_flow_age_spatial.py
```diff
@@ -0,0 +1,116 @@
+#
+# Example simulation script with spatial data output and visualisation
+# Currently running with Kit's gibraltar set-up
```
KCGallagher, line 3: Yes I think let's move this to a sub directory within the Gibraltar simulation, and use the same input and parameters files

python_examples/age_stratified_plot.py
```diff
@@ -0,0 +1,163 @@
...
+        # with the substring "age"
+        self.age_name = next((s for s in list(self.data.columns)
+                             if "age" in s.lower()), None)
+        self.do_ages = (self.age_name is not None)
+        if self.do_ages and (self.age_list is None):
+            num = self.data[self.age_name].max()
+            self.age_list = [str(5*i)+"-"+str(5*i+5) for i in range(num+1)]
+
+    def sum_infectious(self) -> None:
+        """Helper function which sums across all columns containing infectious
```
KCGallagher, line 51: Helper functions should generally be marked as private I think

python_examples/simulation_flow_withage.py
```diff
@@ -0,0 +1,89 @@
...
+# Need to close the writer object at the end of each simulation.
+del (sim.writer)
+del (sim)
+
+# Plotter where age is summed over (to compare to simulation without age)
+logging.getLogger("matplotlib").setLevel(logging.WARNING)
+filename = os.path.join(os.path.dirname(__file__), "simulation_outputs",
+                        "output_withage.csv")
+df = pd.read_csv(filename)
+df_sum_age = df.copy()
```
KCGallagher, line 66: Why do we need an explicit copy here, do we use `df`later? (I couldn't see it used but might have missed it)

KCGallagher (changes requested):
Overall looks good! Will look at some of the knitty gritty of the code separately but some code style changes here.

In particular:
- We try to avoid code that restates what the code does (line by line). The code should be readable enough this isn't necessary (and in most cases it already is readable enough so we can just delete these comments, particularly outside of workflow files).
- Import order should follow PEP8 guidelines - I've sent though this link before but let me know if you can't find it on teams/Google. I've commented/suggested changes in a few places but this should be consistent across all files in the PR
- I think your example is based on the Gibraltar simulation, in which case we could move it to a sub directory of that folder (and use the same parameter/input files instead of duplicating them)
- You've accidentally deleted a number of output files such as the animation gif so if you could restore these that would be much appreciated. I think you might have also duplicated a few other parameters file so maybe check on that too?

Thanks so much!

KCGallagher:
As well as moving the examples into a separate directory, I wonder if we need all the separate example workflows - they mostly seem duplicates of each other? I think maybe we only need the spatial age workflow (based on Gibraltar), and we can remove the other simpler examples - I think adding more example workflows than necessary just makes it more confusing for a new user! Do shout if you've got a strong case for the others though

pyEpiabm/pyEpiabm/core/cell.py
```diff
@@ -96,25 +98,26 @@ def notify_person_status_change(
             Person's old infection status
         new_status : InfectionStatus
             Person's new infection status
-
+        age_group : Age group index
```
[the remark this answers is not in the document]
Elizabeth-Hayman: good idea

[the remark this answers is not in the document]
Elizabeth-Hayman: I've realised that we don't actually pass the person to that function, just some explicit attributes. (As indeed, we do in the notify person status change, the other interaction with the counter). I don't think it would be any easier to pass the person in as an argument instead of just their age, and feels like that would require most info being passed around.

age_stratified_example/sim_outputs/output.csv
```diff
@@ -0,0 +1 @@
+time,age_group,cell,location_x,location_y,InfectionStatus.Susceptible,InfectionStatus.Exposed,InfectionStatus.InfectASympt,InfectionStatus.InfectMild,InfectionStatus.InfectGP,InfectionStatus.InfectHosp,InfectionStatus.InfectICU,InfectionStatus.InfectICURecov,InfectionStatus.Recovered,InfectionStatus.Dead
```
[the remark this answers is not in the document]
Elizabeth-Hayman: I've tidied the python_examples folder so each directory contains a full example, including any output folders. I think its better that way but we can talk if you don't like it

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -9,7 +9,7 @@
 
 
 class InitialisePlaceSweep(AbstractSweep):
-    """Class to update people in the "Place"
+    """Class to initialise people in the "Place"
```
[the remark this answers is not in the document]
Elizabeth-Hayman: It's the initialisation sweep so it puts people in the place. As opposed to the update sweep that takes people out, then pops some different ones back in.

pyEpiabm/pyEpiabm/sweep/update_place_sweep.py
```diff
@@ -43,10 +43,11 @@ def __call__(self, time: float):
                     # Variable population is people not in the fixed pop.
                     # Held in the last group of the place.
```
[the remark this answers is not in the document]
Elizabeth-Hayman: No, that was my bad, getting mixed between dictionary keys and list indexing

pyEpiabm/pyEpiabm/tests/test_unit/test_core/test_microcell.py
```diff
@@ -27,6 +27,7 @@ def test_set_id(self):
         self.assertEqual(self.microcell.id, 2.0)
 
     def test_add_person(self):
+        # Add test to check if age is preserved
```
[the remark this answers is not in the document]
Elizabeth-Hayman: Not sure, but added one anyway

python_examples/age_simulation.py
Elizabeth-Hayman: old file, deleted

python_examples/age_stratified_example/age_stratified_plot.py
Elizabeth-Hayman: :(

python_examples/age_stratified_example/gibraltar_parameters.json
Elizabeth-Hayman: Because my Gibraltar simulation was very much an adaption of Kits, it incorporates all and hence supersedes anything previous. I therefore have just one folder, "Gibraltar example" which should contain everything. The compatibility of the main flow file wil lallow us to turn things on/off as needed

pyEpiabm/pyEpiabm/tests/test_func/test_sim_functional.py
```diff
@@ -139,6 +140,7 @@ def test_population_conservation(self, mock_mkdir, mock_output):
         mock_mkdir.assert_called_with(folder)
         nb_age_group = len(pe.Parameters.instance().age_proportions)
         mock_output_count = iter_num * nb_age_group * cell_count
+        print(mock_output_count)
```
KCGallagher, line 143: Spurious print statement - I guess that was in here for testing, but maybe should be taken out for the final version

python_examples/Gibraltar_example/gibraltar_inputs/microcell_conversion.py
```diff
@@ -0,0 +1,73 @@
+#
+# Converts gibraltar population data from covidsim into compatible input file
+#
```
KCGallagher, line 3: If this a duplicated file?

age_stratified_example/sim_outputs/output.csv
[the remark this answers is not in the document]
KCGallagher: I like that - but does that mean this spurious output file that is not within the `python_examples` directory is no longer needed?

KCGallagher (changes requested):
Thanks for your tidying efforts! I think it would be good to chat about which workflows we need - for example if your gibraltar workflow has superseded mine, then feel free to delete/ replace mine (and associated files). To help this, it would be helpful if you could updated the `python_examples` readme to describe the role of each workflow, so we can work out is there is any duplication.

On the theme of avoiding clutter in this folder, might you be able to double check the plots you've added - I think some are autogenerated with parameter sets that make them meaningless/hard to read, and then this makes it harder to find the really great plots you've been working on!

pyEpiabm/pyEpiabm/core/cell.py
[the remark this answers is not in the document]
KCGallagher: That makes sense - I was just worried about the duplication of passing both but very happy with you not passing the person here if it is not needed!

python_examples/age_stratified_example/gibraltar_parameters.json
KCGallagher: This is great - it sounds like we don't need to old version anymore then?

age_stratified_example/sim_outputs/output.csv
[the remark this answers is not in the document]
Elizabeth-Hayman: There shouldn't be any files on their own now. I'm not seeing any local, please deltete is they're still here

python_examples/Gibraltar_example/gibraltar_inputs/microcell_conversion.py
Elizabeth-Hayman: yes at the moment. Wanted to keep all folders self contained as am anticipating deleting the covdisim comparisonfolder

python_examples/age_stratified_example/gibraltar_parameters.json
Elizabeth-Hayman: Don;t think so... always tentative to delete though XD

Elizabeth-Hayman:
Kit's made the point that there's still a lot of python examples, and some use toy datasets left over  from debugging. I'm not against very simple data/param sets to help learn complicated functions, and could make them tidier if people wanted.

python_examples/Gibraltar_example/gibraltar_inputs/microcell_conversion.py
KCGallagher: Yep that makes sense!

age_stratified_example/sim_outputs/output.csv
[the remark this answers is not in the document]
KCGallagher: File was still there (completely outside pyEpiabm directory) - have deleted now

pyEpiabm/pyEpiabm/core/person.py
```diff
@@ -65,6 +65,9 @@ def set_random_age(self):
         else:
             self.age_group = None
```
rccreswell, line 66:
```suggestion
```

pyEpiabm/pyEpiabm/tests/test_func/test_random_seeds.py
```diff
@@ -219,7 +220,8 @@ def test_random_seed(self, mock_mkdir, mock_write):
         folder = os.path.join(os.getcwd(), self.mock_output_dir)
         mock_mkdir.assert_called_with(folder)
         self.assertEqual(mock_mkdir.call_count, 3)
-
+        print(seed_output)
+        print(comp_output)
```
rccreswell, line 224: Are these still needed?

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -0,0 +1,239 @@
...
+        plt.title(title)
+        plt.gca().legend().set_title('')
+        plt.xlabel("Time")
+        plt.savefig(outfile)
+
+
+if __name__ == '__main__':
+    dirname = os.path.dirname(os.path.abspath(__file__))
+    p = Plotter(os.path.join(os.path.dirname(__file__),
+                "simulation_outputs/output_withage.csv"),
```
rccreswell, line 232:
```suggestion
                "simulation_outputs/output_with_age.csv"),
```
&c
Its not that important, its just withage started to look [funny](https://en.wiktionary.org/wiki/-age#English)

rccreswell (changes requested):

Elizabeth-Hayman:
There's a jupyter notebook in the basic simulation example which is completely equivilent to the python file. Except that is didn't work as was obviously outdated. I this a hangover that can be deleted, or do people like the idea of jupyter notebook examples?

KCGallagher:
> There's a jupyter notebook in the basic simulation example which is completely equivilent to the python file. Except that is didn't work as was obviously outdated. I this a hangover that can be deleted, or do people like the idea of jupyter notebook examples?

No I had been thinking the same so very happy for you to delete that!

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -221,7 +221,7 @@ def barchart(self, outfile: str,
         if param_file:
             title = 'New ' + title
         plt.title(title)
-        plt.gca().legend().set_title('')
+        plt.gca().legend().set_title('Age Group')
         plt.xlabel("Time")
```
rccreswell, line 225: Not important but just realized Time label does not seem to be appearing. Do we need a tight_layout.

rccreswell (approved):

Issue #127, opened 2022-03-30
pyEpiabm - covidsim comparison


KCGallagher:
Comparison simulation to CovidSim - use same parameters etc

CovidSim simulations linked to issue https://github.com/KCGallagher/covid-sim/issues/1#issue-1218559579

PR #139, opened 2022-07-21
Py covidsim

KCGallagher:
Current work replicating CovidSim's simulations of Gibraltar. Reliant on work in #134 for a complete (age-dependant) comparison, but publishing anyway so the record is complete. Overall infection curves show moderate agreement with CovidSim under current parameter set.
Progress towards #127.


I-Bouros (approved):

python_examples/covidsim_comparison/gibraltar_inputs/microcell_conversion.py
```diff
@@ -0,0 +1,70 @@
...
+            if np.linalg.norm([x2-x1, y2-y1]) < sep:
+                sep = np.linalg.norm([x2-x1, y2-y1])
+    return sep
+
+
+pe.Parameters.set_file("python_examples/covidsim_comparison/"
+                       + "gibraltar_parameters.json")
+
+mcell_num = 81
+file_path = "python_examples/covidsim_comparison/gibraltar_inputs/"
```
Elizabeth-Hayman, line 30: Would it be good to do absolute paths, or are we happy each machine will have the same local structure

Elizabeth-Hayman (commented):
Very minor

python_examples/covidsim_comparison/gibraltar_inputs/microcell_conversion.py
KCGallagher: I'm not sure what you mean - relative paths ensures that each machine doesn't need the same local structure (whereas an absolute path would require that)? Do you have an example of what you were thinking?

pyEpiabm/pyEpiabm/routine/toy_population_config.py
```diff
@@ -146,7 +146,8 @@ def add_places(population: Population, place_number: int):
         # in each place.
         for cell in population.cells:
             for microcell in cell.microcells:
-                microcell.add_place(place_number, cell.location,
+                count = math.floor(place_number + random.random())
```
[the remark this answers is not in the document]
KCGallagher: If its an integer you will, this allows you to pass a float as well though which is handy when you want to configure an average number of places per microcell, with random variation between them

python_examples/covidsim_comparison/gibraltar_inputs/microcell_conversion.py
rccreswell, line 30: Maybe Lizzi means we should consider using the `os.path.join(os.path.dirname(__file__), ...` to make it less sensitive to where the file is run from

rccreswell (changes requested):
Before merging, we should edit `epiabm/python_examples/README.md` to include the explanation of the covidsim_comparison directory

python_examples/covidsim_comparison/gibraltar_inputs/microcell_conversion.py
KCGallagher: Makes sense, happy to make that change in the workflows (although I can't see why you wouldn't be running them from the top level), and will also edit the readme!

python_examples/README.md
```diff
@@ -21,3 +21,7 @@ It subsequently saves the results to a .csv file, and plots the infection curve
 We also provide a script (`voronoi_plotting_example.py`) for visualising the spatial distribution of the epidemic, generating cells based on a single location coordinate using [Voronoi Tesselation](https://en.wikipedia.org/wiki/Voronoi_diagram).
 
 ![Voronoi_grid.pt](./spatial_example/spatial_outputs/voronoi_grid_img.png)
+
+## CovidSim Comparision
+
+The `gibraltar.simulation.flow.py` script runs a simulation of the spread of coronavirus across Gibraltar - a country chosen for its small size (and hence ease of simulation) while having complete [simulation parameters](https://github.com/mrc-ide/covid-sim/blob/master/data/admin_units/Gibraltar_admin.txt) available on the repo for [CovidSim](https://github.com/mrc-ide/covid-sim). Microcell-wise locations are not specified directly, and so reconstructed using the `microcell.conversion.py` script to generate the `gib_input.csv` file used in the main simulation flow.
```
rccreswell, line 27:
```suggestion
The `gibraltar_simulation_flow.py` script runs a simulation of the spread of coronavirus across Gibraltar - a country chosen for its small size (and hence ease of simulation) while having complete [simulation parameters](https://github.com/mrc-ide/covid-sim/blob/master/data/admin_units/Gibraltar_admin.txt) available on the repo for [CovidSim](https://github.com/mrc-ide/covid-sim). Microcell-wise locations are not specified directly, and so reconstructed using the `microcell.conversion.py` script to generate the `gib_input.csv` file used in the main simulation flow.
```

rccreswell (approved):
Thanks!

PR #140, opened 2022-07-28
Added Place Sweep and extra python binding functionality

NicholasFan235:


KCGallagher (approved):
Just to confirm, does this also include methods to distribute places across a population, or is it still reliant on importing a population with places from python?

PR #146, opened 2022-09-12
fixed output filename

NicholasFan235:
## Summary

Python writers were incorrectly using filename argument before.
Previously writers were treating filename argument as a file path instead of combining it with the output folder argument.
PR corrects this, folder and filename passed to writers are now correctly used.

## Checklist

## Closing issues


pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_simulation.py
```diff
@@ -58,8 +58,6 @@ def test_configure(self, mock_mkdir):
             self.assertIsInstance(test_sim.population, pe.Population)
             del test_sim.writer
         mo.assert_called_with(filename, 'w')
-        mock_mkdir.assert_called_once_with(os.path.join(os.getcwd(),
-                                           self.file_params["output_dir"]))
```
KCGallagher, line 62: Is this mock no longer being called then? I think if we still have the mock we should check it is being called in the right way?

pyEpiabm/pyEpiabm/output/_csv_writer.py
```diff
@@ -4,13 +4,14 @@
...
 import typing
+import os
 
 from pyEpiabm.output.abstract_reporter import AbstractReporter
 
 
 class _CsvWriter(AbstractReporter):
     def __init__(self, folder: str, filename: str, fieldnames: typing.List,
-                 clear_folder: bool = True):
+                 clear_folder: bool = False):
```
KCGallagher, line 14: If we are changing the default behaviour, can you check this still works with the workflows? In particular this will affect the process of gif generation (which is based on all png files it can find in the output folder, and so may mistakenly include historic images if they are left in the folder - probably best to explicitly clear the folder in this workflow), but also worth checking it is happy to overwrite output files of the same name as well

KCGallagher (commented):
Two small things worth checking, but then should be ready to go

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_simulation.py
NicholasFan235: This is only called if the folder doesn't exist and needs to be created. Was called before because the writers deleted the output folder and recreated them, but I removed this otherwise multiple writers couldn't share the same output folder.

pyEpiabm/pyEpiabm/output/_csv_writer.py
NicholasFan235: Doesn't change functionality of 1st simulation run.
Subsequent simulation runs will now overwrite individual output files instead of deleting and recreating the output folder.

KCGallagher: Makes sense, the only change I can think of is with the pngs so maybe best to at least document this - its only an issue if you also decrease the number of timesteps then rerun the plotter as it won't overwite images from the old simulation I think?

KCGallagher (approved):

PR #154, opened 2022-09-28
Examples

Elizabeth-Hayman:
Some smallish changes to the plot configuration, and plots using raw data from covidsim


python_examples/spatial_example/spatial_simulation_flow.py
```diff
@@ -47,10 +44,13 @@
...
 
 
+# Method to set the seed at the start of the simulation, for reproducibility
+pe.routine.Simulation.set_random_seed(seed=42)
 # sim_ and file_params give details for the running of the simulations and
 # where output should be written to.
 sim_params = {"simulation_start_time": 0, "simulation_end_time": 50,
-              "initial_infected_number": 1, "initial_infect_cell": True}
+              "initial_infected_number": 1, "initial_infect_cell": True,
+              "simulation_seed": 42}
```
KCGallagher, line 53: Duplication on setting the seed

KCGallagher (changes requested):
Nice changes to the plotting!
Can we increase font size on `python_examples/age_stratified_example/simulation_outputs/age_stratify.png` maybe - its even smaller now and very hard to read! Doesn't need to be the size you'd need in a paper but a little bigger!
Was deleting `python_examples/gibraltar_example/gibraltar_inputs/pop_summary.csv` intentional?
And might you be able to revert the changes to non-gibraltar sim examples - I assume these weren't intentional? This is very much the danger of `git add .` - you might find it easier to use the graphical interface in vscode to stage changes on git when working on plotting so that you make sure you only commit the changes you mean to, as its always more work to revert them later!

python_examples/age_stratified_example/age_stratified_plot.py
```diff
@@ -212,21 +215,24 @@ def barchart(self, outfile: str,
...
         if self.sum_weekly:
             title = "Weekly cases by age"
         else:
             title = "Daily cases by age"
             plt.gca().set_xticks(plt.gca().get_xticks()[::3])  # Avoids overlap
         if param_file:
             title = 'New ' + title
-        plt.title(title)
-        plt.gca().legend().set_title('Age Group')
+        # plt.title(title)  # option to add title is required
```
KCGallagher, line 234:
```suggestion
        # plt.title(title)
```

KCGallagher (approved):
You've replaced the spatial examples folder with slightly different files, but not a big deal. One tiny change for a defunct comment but otherwise am happy to approve this!

PR #182, opened 2023-03-01
Household creation

laraherriott:
## Summary

Previously multiple methods for creating households were used, and both called in the case where there were microcells with no households leading to erroneous population duplication. The methods have now been streamlined and microcells with no households are allowed. In addition calling of the function to calculate the number of children has been made more secure.

## Checklist


pyEpiabm/pyEpiabm/core/microcell.py
```diff
@@ -131,9 +132,12 @@ def add_household(self, people: list):
...
         """
-        household = Household(self, loc=self.location)
-        for person in people:
-            household.add_person(person)
+        if len(people) != 0:
+            household = Household(self, loc=self.location)
+            for person in people:
+                household.add_person(person)
+        else:
+            logging.info("Cannot create empty household")
```
KCGallagher, line 140:
```suggestion
            logging.warning("Cannot create empty household")
```

pyEpiabm/pyEpiabm/routine/file_population_config.py
```diff
@@ -175,18 +182,18 @@ def add_households(microcell: Microcell, household_number: int):
         """
         # Initialises another multinomial distribution
```
KCGallagher, line 183: Is this a duplicate of the method in the toy_pop config? Maybe we can import that in here rather than having duplicated code

pyEpiabm/pyEpiabm/sweep/initial_household_sweep.py
```diff
@@ -439,7 +440,7 @@ def __call__(self, sim_params):
                         elif len(household.persons) == 2:
                             self.two_person_household_ages(household.persons)
 
-                        else:
+                        elif len(household.persons) >= 3:
```
KCGallagher, line 443: Is there a final else clause if this isn't hit either? or should that have already hit the valueerror above?

pyEpiabm/pyEpiabm/tests/test_unit/test_property/test_personal_foi.py
```diff
@@ -26,15 +26,14 @@ def setUpClass(cls) -> None:
...
+        self.assertEqual(result, 1)
         self.assertIsInstance(result, float)
-        self.assertEqual(result, self.infector.infectiousness)
 
     def test_person_susc(self):
         result = PersonalInfection.person_susc(self.infector,
                                                self.infectee,
                                                self.time)
-        self.assertTrue(result > 0)
+        self.assertEqual(result, 1.0)
```
KCGallagher, line 36: Tiny thing, but is there a reason why the person_inf is 1 but the person_susc is 1.0?

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
```diff
@@ -23,6 +23,8 @@ def setUp(self) -> None:
     def test_make_pop_no_file(self, mock_log):
         """Tests for when no file is specified.
         """
+        pe.Parameters.instance().household_size_distribution = []
```
KCGallagher, line 26: Is this needed in every test? Could it go in the setUp() instead? (And why do we need it everytime remind me?

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
```diff
@@ -179,23 +190,34 @@ def test_print_population_loc(self, mock_write, mock_read):
         mock_write.assert_called_once()
         self.assertEqual(mock_write.call_args[0], ('output.csv',))
 
+    @patch('logging.info')
```
KCGallagher, line 193:
```suggestion
    @patch('logging.warning')
```
(Because I changed the logging level above)

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
```diff
@@ -179,23 +190,34 @@ def test_print_population_loc(self, mock_write, mock_read):
...
     @patch("pandas.read_csv")
     @patch("copy.copy")
-    def test_print_population(self, mock_copy, mock_read):
+    def test_print_population(self, mock_copy, mock_read, mock_house_log):
```
KCGallagher, line 196: Do we check that the mock_house_log has been called? I missed this if so, but would be good to check

KCGallagher (changes requested):
Overall good. One bigger change, just to make sure we only have one set of code that makes all the households, and are not duplicating this in multiple places!

pyEpiabm/pyEpiabm/routine/file_population_config.py
laraherriott: Its not quite a duplicate. In the file population the method takes a microcell whereas in toy population it takes a population. I could call the method from the file population into toy population at the level of the microcell. Also the way q is written it is always 1 so I wasn't sure if it was worth removing that line altogether?

pyEpiabm/pyEpiabm/sweep/initial_household_sweep.py
```diff
@@ -242,6 +242,11 @@ def calc_number_of_children(self, household_size: int):
 
         n = household_size
         r = random.random()
+
+        if n < 3:
+            raise ValueError("Household must be greater than 3 people for " +
```
rccreswell, line 247:
```suggestion
            raise ValueError("household_size must be greater than 3 people for " +
```

rccreswell (changes requested):

pyEpiabm/pyEpiabm/tests/test_unit/test_property/test_personal_foi.py
laraherriott: Nope, they're both floats so the one is really 1.0. I will change for clarity.

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
laraherriott: Have moved it to the set up. I added since the household_size_distribution it specified in the parameters file, but the way it is now set up this will call the method in InitialHouseholdSweep, so would need mocking every time. Therefore I set the parameter list to empty in all but one test to restrict the tests to using the methods within file population config.

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
laraherriott: This shouldn't have been left in - I have removed.

pyEpiabm/pyEpiabm/sweep/initial_household_sweep.py
```diff
@@ -242,6 +242,11 @@ def calc_number_of_children(self, household_size: int):
 
         n = household_size
         r = random.random()
+
+        if n < 3:
+            raise ValueError("household_size must be greater than 3 people" +
+                             "for calc_number_of_children to be called")
```
KCGallagher, line 248:
```suggestion
            raise ValueError("household_size must be greater than 3 people" +
                             " for calc_number_of_children() to be called")
```

pyEpiabm/pyEpiabm/sweep/initial_household_sweep.py
KCGallagher: Hmm maybe still good to have a check here just incase - add an else clause with a similar logging warning if it encounters an empty household

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
```diff
@@ -67,10 +75,14 @@ def __call__(self, *args):
...
+                    if person_list is not None:
+                        person_list_use = person_list.copy()
+                    else:
+                        person_list_use = None
                     helper.update_place_group(
                         place,
                         group_size=ave_group_size,
-                        person_list=person_list.copy(),
+                        person_list=person_list_use,
                         person_weights=weights,
```
KCGallagher, line 86: I'm slightly what is going on here with all the default commands. What happens if these arguments aren't specified (it looks like no one is added?) When does the `person_list` default to None?

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
```diff
@@ -194,8 +197,17 @@ def test_print_population(self, mock_copy, mock_read):
...
+            expected_df.drop('household_number', axis=1)
+
+            actual_df = mock_copy.call_args.args[0]
+
+            pd.testing.assert_frame_equal(actual_df,
+                                          expected_df, check_dtype=False)
+            household_list = self.df.loc[:, 'household_number']
+            actual_household = actual_df.loc[:, 'household_number']
+            for i in range(len(household_list)):
+                self.assertLessEqual(actual_household[i], household_list[i])
```
KCGallagher, line 210: Why is `assertLessEqual` used here, should the values not just be the same? Also this check can be outside the if clause, as it doesn't require the updated pandas that line 205 does

KCGallagher (changes requested):
All good, just a few small changes and the broken functional test!

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
laraherriott: If no care home parameter values are specified then care homes were being created with the person list from the previous place (because of person_list.copy() and the fact that in the absence of parameters create_age_weights() never gets called). So values will now default when no care home parameters are provided such that no people are added to care homes.

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
laraherriott: Because empty households are no longer allowed the number of households after configuring the population may be lower than the number initial specified, hence LessEqual.

KCGallagher (approved):

pyEpiabm/pyEpiabm/sweep/initial_place_sweep.py
rccreswell: When person_list=None in update_place_group, it looks like it is defaulted to all people in the cell. It looks like it will still work as intended because you also have mean_capacity=0 in that case, but would it be more clear to have person_list_use = [] instead of None?

rccreswell (approved):

PR #190, opened 2023-03-06
Age and age groups

### LINKED ISSUE #189: Match age group and age
**Describe the bug**
The function add_people in the class microcell gives you the option to specify the age group. In line 104 it than assigns an random age when creating the person in which the age group depends on Parameters.instance().age_proportions. In line 108 it sets the age_group based on the provided age_group in the add_people function, but the actual age is never adjusted. So people can be in age group 9 (45-40) and have an age of 6.

**To Reproduce**
Steps to reproduce the behaviour:
1. Set Parameters to...
2. Run with the configuration...

**Expected behaviour**

**Log Files**
If applicable, add log files to explain what happened

**Desktop (please complete the following information):**
 - OS:
 - Python/ C++ Version:

**Additional context**

HenrietteCapel:
## Summary

When adding people to a microcell by the function add_people() a random age was set and afterwards the age group was set to the provided age group (0 if none provided). Therefore age and age_group did not match, and everyone was set to age group 0 if no age_group was specified (while age could still be between 0-85). 
I solved this issue by setting the age group in line with the actual age. Simulations change slightly because before someone could be 20 while having an age group of 15 (75-80 years old), and therefore for example their probability to be asymptomatic was incorrectly. 
Besides, test were adjusted as the compartment counter stored everyone in group 0 before because in most tests a couple of people are added to the test population using the add_people() function without specifying an age. Now the tests also include testing if their age is stored correctly in the compartment counter. The number of people in the test population was reduced to reduce time as now you need to loop over everyone in the population to check their age_group.

## Further work

## Checklist

## Closing issues
closes #189


KCGallagher (approved):
Great spot and great work! I'm looking at passing `age_group=None` to `Person()` - do we generally know the age group when we create a person or would we typically just want to generate a person of any age (which this function ensures is representative of the overall age distribution)?

Overall great work!

HenrietteCapel:
In `toy_population_config.py` we create a population based on the age distribution stored in `Parameters.instance().age_proportions`.

Issue #196, opened 2023-03-08
custom setupclass in functional test


jiayuanz3:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. 

**Describe the solution you'd like**
If it is a new feature, where should it be added?

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or implementations you've considered.

**Comparison to CovidSim**
Give any relevant comparison to the CovidSim code, and describe any deviations from their functionality

**Additional context**

PR #201, opened 2023-03-09
Intervention Tests - Group Comparison Bugfix

### LINKED ISSUE #198: pyEpiabm - Functional test intervention comparisons bug
**Describe the bug**
`compare_susceptible_groups()` method in the test_func.helper_func uses `assertGreaterEqual` which will therefore not pick up cases where the intervention has no effect

**To Reproduce**
Prevent intervention sweeps from running, but the tests on these sweeps still pass

**Expected behaviour**
These tests should fail if the two population groups are completely identical. However we cannot use `assertGreater` as many compartments in this population will be zero - therefore we should only require strictly greater on non-zero compartments. I will write a custom method to implement this behaviour.

**Desktop (please complete the following information):**
All (N/A)

KCGallagher:
## Summary

`compare_susceptible_groups()` method in the test_func.helper_func uses `assertGreaterEqual` which will therefore not pick up cases where the intervention has no effect

These tests should fail if the two population groups are completely identical. However we cannot use `assertGreater` as many compartments in this population will be zero - therefore we should only require strictly greater on non-zero compartments. I have written a custom method to implement this behaviour.

**This PR contains history from #196 so should only be merged after that**
(_Changes in this PR alone are contained in commit [51d8d44657d768b86099c59789a476251dc24801](https://github.com/SABS-R3-Epidemiology/epiabm/pull/201/commits/51d8d44657d768b86099c59789a476251dc24801#diff-be0274639fb7bc317edfcb9bc70ee470d9ee2c7a18215ce5d78ade7c0ae34fd2)_)
## Checklist

## Closing issues

Resolves #198


jiayuanz3 (approved):

pyEpiabm/pyEpiabm/tests/test_func/helper_func.py
```diff
@@ -23,23 +41,23 @@ def compare_susceptible_groups(self, small_pop, large_pop,
...
         small_pop : list
             The population with fewer individuals
-        large_pop: list
+        large_pop : list
             The population with more individuals
-        status: InfectionStatus
+        status : InfectionStatus
             The infection status wants to be compared
-        method: string
+        method : string
```
rccreswell, line 48:
```suggestion
        method : {'greater', 'equal'}
```
not really relevant to this PR but I think this is the correct docstring when there are a few choices it should match

rccreswell (approved):

PR #214, opened 2023-03-16
Speed-up cell set up

### LINKED ISSUE #213: Speed-up cell set up
When reading in the input file we check for every line if the cell is a new cell or similar to the previous considered cell by looping over the whole population and compare cell IDs. This is time consuming.

Sort the input file on cells and store the current cell. Compare current cell ID with the cell ID in the new line and make a new cell if they do not match.

**Describe alternatives you've considered**
We could also store a list of existing cell IDs but the solution above is more efficient. 

**Comparison to CovidSim**
Give any relevant comparison to the CovidSim code, and describe any deviations from their functionality

**Additional context**

HenrietteCapel:
## Summary
When reading in the input file we check for every line if the cell is a new cell or similar to the previous considered cell by looping over the whole population and compare cell IDs. This is time consuming.
Solution: Sort the input file on cells and store the current cell. Compare current cell ID with the cell ID in the new line and make a new cell if they do not match.

## Checklist

## Closing issues

closes #213

### DIFF
--- a/pyEpiabm/pyEpiabm/routine/file_population_config.py
+++ b/pyEpiabm/pyEpiabm/routine/file_population_config.py
@@ -64,7 +64,7 @@ class FilePopulationFactory:
         # Read file into pandas dataframe
         input = pd.read_csv(input_file)
         loc_given = ("location_x" and "location_y" in input.columns.values)
-        # Sort csv on cell id
+        # Sort csv on cell and microcell ID
         input = input.sort_values(by=["cell", "microcell"])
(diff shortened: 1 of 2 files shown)


KCGallagher (changes requested):
Looks great - it would be good to add some tests for this new functionality though! I.e. pass in a mock input file (out of order*) and check that it does get ordered and makes the right number of cells (and associated microcells).

*In both cells and microcells, as you sort by both

PR #216, opened 2023-04-04
Spatial r0 fix

njs59:
## Summary

Solving the problem that we previously had that we have R_0 spatial infections at every timestep.
Now there are only R_0 total spatial infections per infected individual's total infection

## Further work

## Checklist

## Closing issues

### DIFF
--- a/pyEpiabm/pyEpiabm/tests/test_func/helper_func.py
+++ b/pyEpiabm/pyEpiabm/tests/test_func/helper_func.py
@@ -3,6 +3,7 @@
 #
 
 import unittest
+import numpy as np
 
 import pyEpiabm as pe
 from pyEpiabm.property.infection_status import InfectionStatus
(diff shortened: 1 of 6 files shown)


python_examples/basic_simulation/simple_parameters.json
```diff
@@ -1,5 +1,6 @@
 {
     "basic_reproduction_num": 2.8,
+    "average_infection_time": 17,
```
KCGallagher, line 3: I think we no longer need this!

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/tests/test_func/helper_func.py
```diff
@@ -50,25 +51,24 @@ def compare_susceptible_groups(self, small_pop, large_pop,
 
         """
         for age_group in range(len(pe.Parameters.instance().age_proportions)):
-            with self.subTest(age_group=age_group):
-                if method == 'greater':
-                    self.assertNonZeroGreater(
+            if method == 'greater':
+                self.assertNonZeroGreater(
```
KCGallagher, line 55: Are we summing over age groups here? I would prefer to test each age group separately if possible, but if we can't do that then we certainly don't need to iterate over them in L53! 

The gold standard here would be to that each age group category is greater or equal (i.e. no age group is worse off under the intervention), and also check the sum is strictly greater (i.e. there is a net positive impact of the intervention). I can help write this if you have difficulties, but it should be fairly straightforward from replicating content here, and will just need a detailed docstring to explain what you are doing.

KCGallagher (changes requested):
Please update the description of the spatial inf method - you have not done this yet and I think we should have a justification for this more complicated method. 

I am also not happy with how you have changed the functional tests I'm afraid - the new changes don't seem to make sense. The gold standard here would be to that each age group category is greater or equal (i.e. no age group is worse off under the intervention), and also check the sum is strictly greater (i.e. there is a net positive impact of the intervention). I can help write this if you have difficulties, but it should be fairly straightforward from replicating content here, and will just need a detailed docstring to explain what you are doing.

KCGallagher (approved):

PR #217, opened 2023-05-02
check distancing_enhanced is present before checking the value

jiayuanz3:
## Summary

Check distancing_enhanced is present before checking the value. otherwise, attributeerror when running large simulations (NZ).


HenrietteCapel (commented):
LGTM

HenrietteCapel (approved):

PR #218, opened 2023-05-31
Multiple same category intervention modified

### LINKED ISSUE #206: Multiple same category intervention
Same category intervention may occur more than once. For example, we may want place closes from March to June for the first wave and September to November for the second wave if we want longer simulation time.

As JSON inhibits same object name, we can put each intervention's parameters (dictionary) into a list which will affect the current structure in InterventionSweep.

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or implementations you've considered.

**Comparison to CovidSim**
Give any relevant comparison to the CovidSim code, and describe any deviations from their functionality

**Additional context**

jiayuanz3:
## Summary

interventions can be either dictionary (only one such intervention), or list (a sequence of interventions of same type)

## Closing issues

Put `closes #206 ` in your comment to auto-close the issue that your PR fixes (if such).
If agree with this one, delete https://github.com/SABS-R3-Epidemiology/epiabm/pull/215 PR and close the issue

### DIFF
--- a/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
+++ b/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
@@ -48,12 +48,12 @@ class InterventionSweep(AbstractSweep):
 
     def bind_population(self, population):
         self._population = population
-        for intervention_type, intervention_object in self.\
+        for intervention_key, intervention_object in self.\
                 intervention_params.items():
             if isinstance(intervention_object, list):
                 for index, single_object in enumerate(intervention_object):
                     intervention_init = self.intervention_dict[
-                        intervention_type](
+                        intervention_key](
                         population=self._population, **single_object)
                     intervention_init.occurrence_index = index
                     self.intervention_active_status[intervention_init] = False
@@ -63,9 +63,9 @@ class InterventionSweep(AbstractSweep):
                     # with corrpesonding parameters in the __call__ function
                     if index == 0:
                         Parameters.instance().intervention_params[
-                            intervention_type] = single_object
+                            intervention_key] = single_object
             else:
-                intervention_init = self.intervention_dict[intervention_type](
+                intervention_init = self.intervention_dict[intervention_key](
                             population=self._population, **intervention_object)
                 self.intervention_active_status[intervention_init] = False
 
@@ -94,15 +94,12 @@ class InterventionSweep(AbstractSweep):
                     # whose values could be retrieved with respect to index
                     if hasattr(intervention, 'occurrence_index'):
                         # Get the intervention type
-                        intervention_type = \
-                            list(self.intervention_dict.keys())[list(
-                                self.intervention_dict.values()).index(
-                                type(intervention))]
+                        intervention_key = intervention.name
                         # Update parameter values with current
                         # active intervention
                         Parameters.instance().intervention_params[
-                            intervention_type] = self.intervention_params[
-                            intervention_type][
+                            intervention_key] = self.intervention_params[
+                            intervention_key][
                             intervention.occurrence_index]
                     self.intervention_active_status[intervention] = True
                 intervention(time)
--- a/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
+++ b/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
@@ -7,6 +7,7 @@ from pyEpiabm.intervention import CaseIsolation, Vaccination, PlaceClosure
 from pyEpiabm.intervention import HouseholdQuarantine, SocialDistancing
 from pyEpiabm.intervention import DiseaseTesting, TravelIsolation
 from .abstract_sweep import AbstractSweep
+import warnings
 
 
 class InterventionSweep(AbstractSweep):
@@ -57,6 +58,14 @@ class InterventionSweep(AbstractSweep):
                         population=self._population, **single_object)
                     intervention_init.occurrence_index = index
                     self.intervention_active_status[intervention_init] = False
+
+                    # Calculate the current intervention's end date
+                    current_end_date = single_object['start_time'] + \
+                        [value for key, value in single_object.items() if
+                         'delay' in key][0] + \
+                        [value for key, value in single_object.items() if
+                         'duration' in key][0]
+
                     # If there are multiple same class of interventions,
                     # initialise it with the first invervention.
                     # Other interventions will be activated and updated
@@ -64,6 +73,15 @@ class InterventionSweep(AbstractSweep):
                     if index == 0:
                         Parameters.instance().intervention_params[
                             intervention_key] = single_object
+                        previous_end_date = current_end_date
+
+                    # Raise warning message if concurrent interventions occur
+                    else:
+                        if previous_end_date > single_object['start_time']:
+                            warnings.warn(
+                                'Concurrent interventions should not occur!')
+                    previous_end_date = current_end_date
+
             else:
                 intervention_init = self.intervention_dict[intervention_key](
                             population=self._population, **intervention_object)
--- a/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
+++ b/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
@@ -2,12 +2,14 @@
 # Sweeps for taking care of the interventions
 #
 
+import logging
+
 from pyEpiabm.core import Parameters
 from pyEpiabm.intervention import CaseIsolation, Vaccination, PlaceClosure
 from pyEpiabm.intervention import HouseholdQuarantine, SocialDistancing
 from pyEpiabm.intervention import DiseaseTesting, TravelIsolation
+
 from .abstract_sweep import AbstractSweep
-import logging
 
 
 class InterventionSweep(AbstractSweep):
@@ -67,9 +69,9 @@ class InterventionSweep(AbstractSweep):
                          'duration' in key][0]
 
                     # If there are multiple same class of interventions,
-                    # initialise it with the first invervention.
+                    # initialise it with the first intervention.
                     # Other interventions will be activated and updated
-                    # with corrpesonding parameters in the __call__ function
+                    # with corresponding parameters in the __call__ function
                     if index == 0:
                         Parameters.instance().intervention_params[
                             intervention_key] = single_object
@@ -100,7 +102,7 @@ class InterventionSweep(AbstractSweep):
         for intervention in self.intervention_active_status.keys():
             # TODO:
             # - Include an alternative way of case-count.
-            #   Idealy this will be a global parameter that we can plot
+            #   Ideally this will be a global parameter that we can plot
             # - Include condition on ICU
             #   Intervention will be activated based on time and cases now.
             #   We would like to implement a threshold based on ICU numbers.
@@ -124,6 +126,6 @@ class InterventionSweep(AbstractSweep):
                 intervention(time)
 
             elif self.intervention_active_status[intervention] is True:
-                # turn off intervention
+                # Turn off intervention
                 self.intervention_active_status[intervention] = False
                 intervention.turn_off()
--- a/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
+++ b/pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
@@ -48,23 +48,25 @@ class InterventionSweep(AbstractSweep):
 
     def bind_population(self, population):
         self._population = population
-        for intervention in self.intervention_params.keys():
-            if isinstance(self.intervention_params[intervention],
-                          list):
-                for index in range(len(self.intervention_params[
-                        intervention])):
-                    params = self.intervention_params[intervention][index]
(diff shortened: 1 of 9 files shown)


pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -36,22 +36,36 @@ def __init__(self):
...
         for intervention in self.intervention_params.keys():
-            params = self.intervention_params[intervention]
-            self.intervention_active_status[(intervention_dict[intervention](
-                population=self._population, **params))] = False
+            if isinstance(self.intervention_params[intervention],
+                          list):
+                for index in range(len(self.intervention_params[
+                        intervention])):
+                    params = self.intervention_params[intervention][index]
+                    intervention_init = self.intervention_dict[intervention](
```
KCGallagher, line 57: Can you rename this variable to make it clear it is the intervention class that you are initialising here, rather than the associated sweep you are calling

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -36,22 +36,36 @@ def __init__(self):
...
-
         for intervention in self.intervention_params.keys():
-            params = self.intervention_params[intervention]
-            self.intervention_active_status[(intervention_dict[intervention](
-                population=self._population, **params))] = False
+            if isinstance(self.intervention_params[intervention],
+                          list):
+                for index in range(len(self.intervention_params[
+                        intervention])):
+                    params = self.intervention_params[intervention][index]
```
KCGallagher, line 56: This construction is generally worth avoiding where possible. In this context, you want to access both the index and the values of this list, so it is worth using [`enumerate()`](https://www.geeksforgeeks.org/enumerate-in-python/).

```suggestion
                for index, params in enumerate(self.intervention_params[
                        intervention]):
```

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -36,22 +36,36 @@ def __init__(self):
...
+                    intervention_init = self.intervention_dict[intervention](
+                            population=self._population, **params)
+                    intervention_init.occurrence_index = index
+                    self.intervention_active_status[intervention_init] = False
+                    if index == 0:
```
KCGallagher, line 61: This could do with a comment why you are modifying the original parameters object, and for one index only. Once this intervention occurs, do you update the parameters object to store the details of the next intervention in the list. Can the interventions be concurrent and do you check for this if not?

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -72,9 +86,20 @@ def __call__(self, time):
             num_cases = sum(map(lambda cell: cell.number_infectious(),
                             self._population.cells))
             if intervention.is_active(time, num_cases):
-                intervention(time)
                 if self.intervention_active_status[intervention] is False:
+                    # Update with the single intervention
```
KCGallagher, line 90: Can you make this more informative? What are you updating here? And what information are you adding?

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -72,9 +86,20 @@ def __call__(self, time):
...
+                    # (with respect to index)
+                    if hasattr(intervention, 'occurrence_index'):
+                        intervention_type = \
```
KCGallagher, line 93: This is very unwieldy - could the `intervention` object store its own type directly?

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -72,9 +86,20 @@ def __call__(self, time):
...
+                            list(self.intervention_dict.keys())[list(
+                                self.intervention_dict.values()).index(
+                                type(intervention))]
+                        Parameters.instance().intervention_params[
```
KCGallagher, line 97: Similarly I think these lines are very long and not particularly readable - I'm thinking of ways to tidy it up but for now it would be good to comment what this does if possible

pyEpiabm/pyEpiabm/tests/test_unit/test_sweep/test_intervention_sweep.py
```diff
@@ -113,6 +129,14 @@ def test___call__(self):
                  if isinstance(key, CaseIsolation)][0]])
         self.assertIsNone(self.person_symp.isolation_start_time)
 
+        # The second place closure is active after its starting time
```
KCGallagher, line 132: Can we also check the first place closure is no longer active?

KCGallagher (changes requested):
Mostly good - I added a few suggestions, but I'm afraid I also had a query or two that was more serious. The tests you have are nice but I think it should be more comprehensive. For example, do we allow interventions that overlap in time period? If so, we should test this. If not, then we should handle this explicitly.

KCGallagher (changes requested):
Mostly good - I added a few suggestions, but I'm afraid I also had a query or two that was more serious. The tests you have are nice but I think it should be more comprehensive. For example, do we allow interventions that overlap in time period? If so, we should test this. If not, then we should handle this explicitly.

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -36,22 +36,36 @@ def __init__(self):
...
         self._population = population
-        intervention_dict = {'case_isolation': CaseIsolation,
-                             'place_closure': PlaceClosure,
-                             'household_quarantine': HouseholdQuarantine,
-                             'social_distancing': SocialDistancing,
-                             'disease_testing': DiseaseTesting,
-                             'vaccine_params': Vaccination,
-                             'travel_isolation': TravelIsolation}
-
         for intervention in self.intervention_params.keys():
```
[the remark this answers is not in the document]
jiayuanz3: Done.

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
jiayuanz3: I've added the a comment in the code. Each type of interventions are assumed to be in a sequence, not concurrent and when such intervention is activated, the parameter will be updated. **Concurrent** could handle by the input json file. For instance: from day1-day10 only close School, from day 5-15 only close outdoor. Then parameter will be given as: start_date = 1, duration = 4, close_type = [school]; start_date = 5, duration = 5, close_type = [school, outdoor]; start_date = 10, duration = 5, close_type = [outdoor].

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
jiayuanz3: This hasattr() function checks this active intervention has 'occurrence_index' which means it has several same type of interventions. We'll only update the parameter file when this intervention is active and its intervention type in the original parameter json file is stored as a list (a sequence of same type of interventions). When updating, we need its intervention_type (e.g. CaseIsolation etc.). So I think this batch of code is necessary, that I can't come up with a better way.

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
KCGallagher: > I've added the a comment in the code. Each type of interventions are assumed to be in a sequence, not concurrent and when such intervention is activated, the parameter will be updated. **Concurrent** could handle by the input json file. For instance: from day1-day10 only close School, from day 5-15 only close outdoor. Then parameter will be given as: start_date = 1, duration = 4, close_type = [school]; start_date = 5, duration = 5, close_type = [school, outdoor]; start_date = 10, duration = 5, close_type = [outdoor].

I'm not quite sure what you mean that "Concurrent could handle by the input json file". I think if we are okay with concurrent inputs in the json then we need a test to check that they work as expected. Otherwise we need to raise an error somewhere (or even just flag a warning saying the second one is ignored or similar). I am happy with you allowing concurrent inputs, but I do think this is a tricky case for the code, and so I think we need to test that concurrent interventions
 can be active simultaneously, activated and deactivated separately (at the correct times), and have a summed effect when both are active.

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
KCGallagher: > This hasattr() function checks this active intervention has 'occurrence_index' which means it has several same type of interventions. We'll only update the parameter file when this intervention is active and its intervention type in the original parameter json file is stored as a list (a sequence of same type of interventions). When updating, we need its intervention_type (e.g. CaseIsolation etc.). So I think this batch of code is necessary, that I can't come up with a better way.

So what is the difference between `intervention_type` and `type(intervention)`? I would have expected these to be the same (eliminating the complexity in L98-99), so if they're not then I think we need to think about renaming the `intervention_type` variable to something more accurate, or even replacing it with the type of the intervention object

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
```diff
@@ -87,13 +89,16 @@ def __call__(self, time):
...
+                    # If the next same type intervention is activated,
+                    # update the parameter values with this new intervention
+                    # whose values could be retrieved with respect to index
                     if hasattr(intervention, 'occurrence_index'):
+                        # Get the intervention type
                         intervention_type = \
                             list(self.intervention_dict.keys())[list(
                                 self.intervention_dict.values()).index(
                                 type(intervention))]
+                        # Update parameter values
```
KCGallagher, line 101:
```suggestion
                        # Update parameter values with current active intervention
```

Would this be an accurate description of what is happening? And apologies in advance if this exceeds line length for flake8

pyEpiabm/pyEpiabm/tests/test_unit/test_sweep/test_intervention_sweep.py
```diff
@@ -129,8 +129,15 @@ def test___call__(self):
                  if isinstance(key, CaseIsolation)][0]])
         self.assertIsNone(self.person_symp.isolation_start_time)
 
-        # The second place closure is active after its starting time
         self.interventionsweep(time=200)
+        # The first place closure is inactive after
```
KCGallagher, line 133: Am I misreading this test? It looks like you are asserting that the active status is true, but the comment saying you are checking that the first intervention is now inactive, so I would have expected `assertFalse` here

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
jiayuanz3: intervention_type is a string / key in the json file (e.g. 'place_closure', 'case_isolation), we use this to access / modify the value in parameter json file. type(intervention) is the pre-defined class, which we are trying to initialise (e.g. CaseIsolation(AbstractIntervention))

pyEpiabm/pyEpiabm/tests/test_unit/test_sweep/test_intervention_sweep.py
jiayuanz3: Changed now

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
jiayuanz3: Done!

KCGallagher (changes requested):
I have replies to previous threads with comments - hopefully this addresses all the areas you were confused by?

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
jiayuanz3: Done!

pyEpiabm/pyEpiabm/sweep/intervention_sweep.py
jiayuanz3: From current code, the intervention will last for 15 days, but parameters would be changed from day 5 (intervention_para1 from day1 to day4, intervention_para2 from day5 to day 15). Now I raised such error message by checking whether the previous same_category_intervention'd end date is later than the current intervention's start date.

KCGallagher (changes requested):
This is almost there! The changes are great, but I am afraid the warning needs to be thrown through the logging module to be handled correctly - see other [examples](https://github.com/SABS-R3-Epidemiology/epiabm/blob/0a400a844fd92d30db73cd2feb3c9823f4e69e43/pyEpiabm/pyEpiabm/sweep/initial_infected_sweep.py#L43) in the code.

You then test by mocking the warning function as [so](https://github.com/SABS-R3-Epidemiology/epiabm/blob/0a400a844fd92d30db73cd2feb3c9823f4e69e43/pyEpiabm/pyEpiabm/tests/test_unit/test_sweep/test_initial_infected_sweep.py#L133). 

Otherwise these warnings will output to terminal but may not be saved in the log files. Thanks so much for your effort on this, and this should now be an easy change!

KCGallagher (approved):

PR #219, opened 2023-09-25
Cpp gibraltar

KCGallagher:
Historic Gibraltar simulation code.

Mostly changes the functions which sample from the population. Samples from the population with replacement whereas old code was without replacement, for consistency with pyEpiabm.

Also adds the updated C++ parameters file used for simulations in the original paper, and an example workbook for generating a C++ population object out of a python one.


cEpiabm/examples/basic_example/main.cpp
```diff
@@ -69,17 +69,17 @@ void run()
 
 int main()
 {
-    try
-    {
+    //try
+    //{
         run();
```
rccreswell, line 74: Could we keep just this line and remove all the comments?

rccreswell (changes requested):

cEpiabm/examples/basic_example/main.cpp
KCGallagher: I think it was left as an example, but happy to remove now we have logging for cEpiabm too

rccreswell (approved):

PR #221, opened 2023-09-25
Plotting cpp gibraltar

KCGallagher:


KCGallagher:
Temporary close while work is still under review, in case additional modifications from this branch are required

PR #226, opened 2023-09-26
Place closure infectee

HenrietteCapel:
## Summary
For both the place closure (PC) and social distancing (SD) interventions the susceptibility of the infector was scaled. This should be the susceptibility of the infectee. I have changed this and updated the unit tests. Besides, I removed the infector parameter from the susceptibility function in the place_foi.py and spatial_foi.py file because the parameter was never used.

## Further work

## Checklist

## Closing issues


pyEpiabm/pyEpiabm/property/place_foi.py
```diff
@@ -55,16 +55,14 @@ def place_inf(place, infector, time: float):
         return place_inf
 
     @staticmethod
-    def place_susc(place, infector, infectee,
+    def place_susc(place, infectee,
                    time: float):
```
KCGallagher, line 59:
```suggestion
    def place_susc(place, infectee, time: float):
```

No need to split over lines now I think, but tiny change!

KCGallagher (changes requested):
Maybe in the foi/ spatial_susc methods that you have changed, it would be good to clarify in the method that the closure parameters are based on the properties of the infectee's microcell (this is something we should have documented before, but didn't)

Are the curve's definitely labeled right in `python_examples/intervention_example/intervention_outputs/place_closure_closure_household_infectiousness_Icurve_plot.png` - I would have expected the orange curve to be in the middle (from parameter values), not the green. If they are, do you have any idea why this might be happening (i.e. why there is a turning point in this parameter), and maybe you could document it in the readme specific to these interventions

PR #228, opened 2023-09-26
Toy interventions

### LINKED ISSUE #227: Spurious Reinfection
Recovered individuals may be reinfected if they are selected as part of a reinfection event. Given these interactions may still occur organically, it feels more natural to allow such interactions, but prevent recovered individuals from being queued.

KCGallagher:
Recovered individuals may be reinfected if they are selected as part of a reinfection event. Given these interactions may still occur organically, it feels more natural to allow such interactions, but prevent recovered individuals from being queued. In the commit history, I have also included a person identifier used in debugging - this is only valid for spatial simulations (as it relies on each cell having a unique location) and so is not to be included in main at this point.

Closes #227


KCGallagher:
(Github really should allow PRs to be deleted - this was simply the wrong branch)

PR #229, opened 2023-09-26
Bugfix - recovered reinfection

KCGallagher:
Recovered individuals may be reinfected if they are selected as part of a reinfection event. Given these interactions may still occur organically, it feels more natural to allow such interactions, but prevent recovered individuals from being queued. In the commit history, I have also included a person identifier used in debugging - this is only valid for spatial simulations (as it relies on each cell having a unique location) and so is not to be included in main at this point.

Closes https://github.com/SABS-R3-Epidemiology/epiabm/issues/227

### DIFF
--- a/pyEpiabm/pyEpiabm/core/cell.py
+++ b/pyEpiabm/pyEpiabm/core/cell.py
@@ -82,7 +82,8 @@ class Cell:
         self.id = id
 
     def enqueue_person(self, person: Person):
-        """Add person to queue for processing at end of iteration.
+        """Add person to queue for processing at end of iteration, provided
+        they are not already recovered (and so may be infected).
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


pyEpiabm/pyEpiabm/core/cell.py
```diff
@@ -90,7 +90,8 @@ def enqueue_person(self, person: Person):
             Person to enqueue
 
         """
-        self.person_queue.put(person)
+        if person.infection_status != InfectionStatus.Recovered:
```
rccreswell, line 93: Agreed, just need to note this restriction in the docstring since its not always obvious from the name of the method

rccreswell (changes requested):

rccreswell (approved):

PR #232, opened 2023-09-28
NZ simulation params

jiayuanz3:
## Summary

New Zealand parameters to run the simulation.


python_examples/NZ_example/NZ_flow.py
```diff
@@ -0,0 +1,118 @@
+#
+# Example simulation script running with Bermuda parameters
```
KCGallagher, line 2:
```suggestion
# Example simulation script running with New Zealand parameters
```

KCGallagher (changes requested):
That New Zealand input file is massive! I think if we have the script to generate this (or instructions and a link to epigeopop) then we don't need to add this as well - you could put it in the `.gitignore` instead so you don't have to delete it off your local machine

KCGallagher (approved):

PR #233, opened 2023-09-28
Ber lux sim fix

njs59:
## Summary

Run simulations for bermuda but mainly luxembourg. Included in this are our plots for the paper for the luxembourg sections

## Further work

May need to resize plots for paper

## Checklist

## Closing issues


KCGallagher (commented):
Good work @njs59 ! Will let you fix issues in the morning before doing a full review, but just to reiterate a few points we mentioned this afternoon - we want to avoid large csv files being committed (say 2000+ lines as a reference, certainly nothing over 10,000). we also want to avoid code duplication - I understand you used a number of different simulation scripts when trying out ideas, but we don't need to commit these all now - only scripts that have markedly different functionality, and are used in the paper, should be committed.

As a guide, Henriette has done a great job on the interventions examples - duplicate scripts have been removed, and workflows consolidated. Each file is clearly named to demonstrate its individual functionality, and the readme clearly describes what each file does (with one file per intervention). In general, I would say one workflow per country would be a good guide (or two if you want one with and without interventions), each separately documented with example outputs in the readme.

@laraherriott has suggested it might be easier to create a separate branch off main, and then just copy across the Luxembourg simulation script and the relevant notebook - this would take about 5 minutes to complete so would be much faster than undoing all the changes here, and sorting through all the different code files.

KCGallagher:
Superseded by #235, which doesn't have the accidental, test-breaking changes, or tinkering associated with investigation into the impact of interventions

PR #236, opened 2023-09-29
gibraltar with intervention results

jiayuanz3:
## Summary

Gibraltar example of with and without interventions.


KCGallagher (approved):
This is great! Could we maybe add a little readme that just shows the infection curves before and after, with a short narrative of the changes? Doesn't need to be long as all, but would just explain what is going on here. Happy to pre-approve conditional on you doing that though, in case I'm in meetings when you finish

PR #239, opened 2023-10-02
Rt inference

laraherriott:
## Summary

Notebook to illustrate inference of Rt on the New Zealand output files.


python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+    "import plotly.graph_objs as go"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "6b8e0ebf",
+   "metadata": {},
+   "source": [
+    "The main parameter required by these models is the serial interval. Since we aim for an overall effective R0 of approximately 3, we searched the literature for information on the serial interval distribution for the original SARS-CoV-2 variant. Estimates of the mean serial interval for this period range from 4.2 to 7.5. On the basis of its use in Menendez (2020), I proceed with an estimated gaussian serial distribution with mean = 4.7 and standard deviation = 2.9 when using fixed serial intervals."
```
rccreswell, line 44:
```suggestion
    "The main parameter required by these models is the serial interval. Since we aim for an overall effective R0 of approximately 3, we searched the literature for information on the serial interval distribution for the original SARS-CoV-2 variant. ....."
```
Looking later in the code, it seems like we are actually using the serial intervals that are read in from the file, rather than the one constructed based on the mean and sd?
Probably easier to just say here that you used the Covid serial intervals which were taken from here (https://github.com/SABS-R3-Epidemiology/transmission-heterogeneity-results/tree/main/data_library/serial_interval)

rccreswell, line 44:
```suggestion
    "The main input parameter required by these models is the serial interval. Since we aim for an overall effective R0 of approximately 3, we searched the literature for information on the serial interval distribution for the original SARS-CoV-2 variant. Estimates of the mean serial interval for this period range from 4.2 to 7.5. On the basis of its use in Menendez (2020), I proceed with an estimated gaussian serial distribution with mean = 4.7 and standard deviation = 2.9 when using fixed serial intervals."
```
Just to clarify that the serial interval is what we are setting in advance, whereas R_t is the main parameter of the model that we are trying to learn from the data

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "2a92b1cb",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "SI_mean = 4.7\n",
+    "SI_sd = 2.9"
```
rccreswell, line 55: If this code is not used later on, let's remove it altogether
I think I originally sent you an older version of this notebook with lots of extraneous code in it, so sorry for being more pedantic now !

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+    "num_timepoints = len(locally_infected_cases)\n",
+    "ws_mean = SI_mean\n",
+    "ws_var = SI_sd**2\n",
+    "theta = ws_var / ws_mean\n",
+    "k = ws_mean / theta\n",
+    "w_dist = scipy.stats.gamma(k, scale=theta)\n",
+    "disc_w = w_dist.pdf(np.arange(num_timepoints))\n",
+    "\n",
+    "\n",
+    "# Same inference, but using the LocImpBranchProPosterior\n",
```
rccreswell, line 133:
```suggestion
    "# Set parameters of the prior on Rt and the sliding window\n",
```

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+   "execution_count": 7,
+   "id": "d37630e8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "x = df['InfectionStatus.Susceptible']\n",
+    "# first generate a list of infected cases\n",
+    "locally_infected_cases = [0] + list(-np.diff(x.values)[:30])\n",
+    "# then create a list of the number of imported cases per day\n",
+    "imported = [1]*len(locally_infected_cases)\n",
```
rccreswell, line 121: Needs a comment saying that this corresponds to new infecteds being introduced during the NZ simulation

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+    "\n",
+    "print('Average Rt over the first week = ', statistics.mean(first_week))\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "43c8d4f7",
+   "metadata": {},
+   "source": [
+    "We then repeat the inference on the number of infections under more strict iterventions."
```
rccreswell, line 1287:
```suggestion
    "We then repeat the inference on the number of infections under more strict interventions."
```

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+    "ws_mean = SI_mean\n",
+    "ws_var = SI_sd**2\n",
+    "theta = ws_var / ws_mean\n",
+    "k = ws_mean / theta\n",
+    "w_dist = scipy.stats.gamma(k, scale=theta)\n",
+    "disc_w = w_dist.pdf(np.arange(num_timepoints))\n",
+    "\n",
+    "\n",
+    "# Same inference, but using the LocImpBranchProPosterior\n",
+    "tau = 3\n",
```
rccreswell, line 134: How sensitive are the results to this parameter value? 
I'm not suggesting that we change away from tau=3, just that if it is causing a significant difference we ought to be more careful about stating upfront (here, and more importantly in the supplement of the paper) that our results are conditional on a choice we decided to make for the sliding window, and maybe justifying this choice in some way. Especially since tau=3 is not necessarily a common choice that is frequently appearing in past applications of this method.

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+    "# Build the serial interval w_s\n",
+    "num_timepoints = len(locally_infected_cases)\n",
+    "ws_mean = SI_mean\n",
+    "ws_var = SI_sd**2\n",
+    "theta = ws_var / ws_mean\n",
+    "k = ws_mean / theta\n",
+    "w_dist = scipy.stats.gamma(k, scale=theta)\n",
+    "disc_w = w_dist.pdf(np.arange(num_timepoints))\n",
+    "\n",
+    "\n",
```
I-Bouros, line 132:
```suggestion
```

Delete these lines as you would only use that if you were generating data during the notebook. But since you read them from a file, there is no need for this block (which you do not even use)

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+    "# Build the serial interval w_s\n",
+    "num_timepoints_strict = len(locally_infected_cases_strict)\n",
+    "ws_mean = SI_mean\n",
+    "ws_var = SI_sd**2\n",
+    "theta = ws_var / ws_mean\n",
+    "k = ws_mean / theta\n",
+    "w_dist = scipy.stats.gamma(k, scale=theta)\n",
+    "disc_w = w_dist.pdf(np.arange(num_timepoints_strict))\n",
+    "\n",
+    "\n",
```
I-Bouros, line 1326: Idem
```suggestion
```

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+    "trace2 = go.Scatter(\n",
+    "    x=list(intervals_strict['Time Points']) + list(intervals_strict['Time Points'])[::-1],\n",
+    "    y=list(intervals_strict['Lower bound CI']) + list(intervals_strict['Upper bound CI'])[::-1],\n",
+    "    fill='toself',\n",
+    "    fillcolor='indigo',\n",
+    "    line_color='indigo',\n",
+    "    opacity=0.5,\n",
+    "    mode='lines',\n",
+    "    name='Credible interval 0.95',\n",
```
I-Bouros, line 2445:
```suggestion
    "    name='Credible interval 95%',\n",
```
I would prefer using percentages, but not too important.

python_examples/R0_inference/learn_R.ipynb
```diff
@@ -0,0 +1,2500 @@
...
+         ]
+        },
+        {
+         "fill": "toself",
+         "fillcolor": "indigo",
+         "line": {
+          "color": "indigo"
+         },
+         "mode": "lines",
+         "name": "Credible interval 0.95",
```
I-Bouros, line 1449: Prefer it as a percentage
```suggestion
         "name": "Credible interval 95%",
```

I-Bouros (changes requested):
I agree with Richard's suggestions. Here are a couple of things I would also suggest you delete / change in addition.

KCGallagher (approved):
Happy with this, and the more standard $\tau = 7 days$

rccreswell (approved):

I-Bouros (approved):

PR #246, opened 2023-11-21
Add base citation file (#245)

### LINKED ISSUE #245: Citation file
Add citation file to integrate with github for ease of use.

See [this](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) docs page for info.

KCGallagher:
Added citation file for github compatibility.
Based on zenodo repo v1.1.0, but also includes link to arxiv.
May add JOSS paper once published

Closes #245

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,67 @@
+# CITATION.cff file for Epiabm, generated with cffinit.
+
+cff-version: 1.1.0
+title: Epiabm
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - given-names: 'Kit '
+    family-names: Gallagher
+    email: gallagher@maths.ox.ac.uk
+    affiliation: 'Mathematical Institute, University of Oxford'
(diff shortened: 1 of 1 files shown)


laraherriott (approved):
Happy to leave as is but we could also update the affiliations from my year to be the current departments in line with those for everyone else? Also the author order seems a bit random (though you may have had reasons to do it in this way) - I'm not sure how much it matters but these files can be pulled into Zotero which would import the author order as given here.

I-Bouros (approved):

KCGallagher:
> > Happy to leave as is but we could also update the affiliations from my year to be the current departments in line with those for everyone else? Also the author order seems a bit random (though you may have had reasons to do it in this way) - I'm not sure how much it matters but these files can be pulled into Zotero which would import the author order as given here.
> 
> I think it reflects the order of contribution on the main code page. I am happy as well to leave as is or change it.

This is exactly the case! I am not sure how else to do it - we have two papers but they do not have the same authorship list/order, so there is no easy way to come up with one consistent order. Happy to hear alternative suggestions here though?

PR #247, opened 2023-11-22
Feature

### LINKED ISSUE #242: ID implementation
Unique identification is needed for both Person and Household in order for storage of infection history.

mghosh00:
## Summary

Created ids for Person and Household, and changed the Cell and Microcell ids to be strings

## Checklist

## Closing issues

Closes #242 .


pyEpiabm/pyEpiabm/core/cell.py
```diff
@@ -6,12 +6,13 @@
 import numpy as np
 from queue import Queue
 from numbers import Number
+import re
```
KCGallagher, line 9:
```suggestion
import numpy as np
import re
from queue import Queue
from numbers import Number
```

Prefer module imports before class imports

pyEpiabm/pyEpiabm/core/cell.py
```diff
@@ -70,16 +71,24 @@ def add_microcells(self, n):
...
+        id : str
             Identity of cell
 
         """
-        self.id = id
+        # Ensure id is a string
+        if not isinstance(id, str):
+            raise TypeError("id must be of type string")
+
+        # May want to set upper limit on the number of digits
```
KCGallagher, line 87: Really comment should match the code. We occasionally allow TODO comments, but I don't think this merits that (should be for changes that resolve active problems) - if you think this is important enough to change then either do so in this PR, or make an issue for it.

KCGallagher: Instead - I would document the regex as this is not readable for everyone

pyEpiabm/pyEpiabm/core/cell.py
```diff
@@ -70,16 +71,24 @@ def add_microcells(self, n):
...
+        if re.match("^\\d+$", id):
+            self.id = id
+        else:
+            raise ValueError("id must take the correct form")
```
KCGallagher, line 91: This isn't very informative - maybe use an f-string to quote the ID provided and then say what the correct form is (all digits)

pyEpiabm/pyEpiabm/core/household.py
```diff
@@ -104,3 +106,26 @@ def remove_household(self):
...
+        Parameters
+        ----------
+        id : str
+            Identity of household
+
+        """
+
+        # Ensure id is a string
+        if not isinstance(id, str):
+            raise TypeError("id must be of type string")
```
KCGallagher, line 125:
```suggestion
            raise TypeError("Provided id must be a string")
```

I think this phrasing is more natural - change throughout

pyEpiabm/pyEpiabm/core/household.py
```diff
@@ -104,3 +106,26 @@ def remove_household(self):
...
+
+        # May want to set upper limit on the number of digits
```
KCGallagher, line 127: Definitely comment the regex match here instead of the vague suggestion - this is much less clear what you are looking for

pyEpiabm/pyEpiabm/core/household.py
```diff
@@ -104,3 +106,26 @@ def remove_household(self):
         """
         self.microcell.households.remove(self)
         self.cell.households.remove(self)
+
+    def set_id(self, id: str):
+        """Updates ID of household (i.e. for input from file).
```
KCGallagher, line 111: Style point - you are referring to a single household so use the appropriate article - otherwise you are referring to an entire class (Cell, Microcell and Household), which are always capitalized in the docs

pyEpiabm/pyEpiabm/core/person.py
```diff
@@ -214,12 +217,35 @@ def remove_person(self):
...
                                    self.age_group)
         self.microcell.cell.persons.remove(self)
         self.microcell.persons.remove(self)
         self.household.persons.remove(self)
+
+    def set_id(self, id: str):
+        """Updates ID of person (i.e. for input from file).
+        ID format: 4.3.2.1 represents cell 4, microcell 3 within this cell,
+        household 2 within this microcell, and person 1 within this
+        household. The id will only be changed if there is a match.
```
KCGallagher, line 234:
```suggestion
        household. The id will only be changed if there is a duplicate.
```


Apologies if this suggested change takes you over line length limits

pyEpiabm/pyEpiabm/routine/file_population_config.py
```diff
@@ -84,8 +84,12 @@ def make_pop(input_file: str, random_seed: int = None, time: float = 0):
         current_cell = None
         # Iterate through lines (one per microcell)
         for _, line in input.iterrows():
+
+            # Converting from float to string
+            cell_id_csv = str(int(line["cell"]))
```
KCGallagher, line 89: Is it possible to specify the type of the columns when we read this csv file in with pandas? Or is there a reason for doing this explicit conversion instead which I've missed?

pyEpiabm/pyEpiabm/routine/file_population_config.py
```diff
@@ -96,13 +100,21 @@ def make_pop(input_file: str, random_seed: int = None, time: float = 0):
...
             # Raise error if microcell exists, then create new one
             for microcell in cell.microcells:
-                if microcell.id == line["microcell"]:
+
+                # The microcell id is a string separated by periods, so we
+                # must find the part which matches the csv input
+                id_parts = microcell.id.split(".")
+                if id_parts[-1] == microcell_id_csv:
                     raise ValueError(f"Duplicate microcells {microcell.id}"
                                      + f" in cell {cell.id}")
```
KCGallagher, line 109: Are we just iterating over the the microcells to check whether the new one is a duplicate off an existing one? If so, it would be much simpler to generate a list of IDs `[microcell.id for microcell in cell.microcells]` and then logically test if the proposed ID is already in this list. Let me know if you need help with this test. It will also be faster than this explicit list.

pyEpiabm/pyEpiabm/routine/file_population_config.py
```diff
@@ -96,13 +100,21 @@ def make_pop(input_file: str, random_seed: int = None, time: float = 0):
...
 
             new_microcell = Microcell(cell)
             cell.microcells.append(new_microcell)
-            new_microcell.set_id(line["microcell"])
+
+            # Either keep the following line (sets id according to data frame
+            # input), or remove it (sets id according to current number of
+            # microcells in cell.microcells)
```
KCGallagher, line 116: I don't understand this comment - are you suggesting we should comment the following line in some contexts? I would be against this, can we automate this different behavior in different contexts

pyEpiabm/pyEpiabm/routine/file_population_config.py
```diff
@@ -150,7 +162,7 @@ def make_pop(input_file: str, random_seed: int = None, time: float = 0):
         return new_pop
 
     @staticmethod
-    def find_cell(population: Population, cell_id: float, current_cell: Cell):
+    def find_cell(population: Population, cell_id: str, current_cell: Cell):
         """Returns cell with given ID in population, creates one if
```
KCGallagher, line 166: Old documentation refers to 'ID', whereas new comments use 'id'. I could be convinced by the new method, as it matches the parameter name, but we need to be consistent.

pyEpiabm/pyEpiabm/core/cell.py
```diff
@@ -70,16 +71,24 @@ def add_microcells(self, n):
         for _ in range(n):
             self.microcells.append(Microcell(self))
 
-    def set_id(self, id: float):
+    def set_id(self, id: str):
```
KCGallagher, line 74: Open question - should we check for uniqueness in this method? It would not be a big time cost, so I think it might make sense to check there is not a duplicate ID in the existing list of cell IDs. This would likely throw an error (or at least be logged) if it is a duplicate. You will also need to test this behavior I'm afraid (and replicate for other classes)

pyEpiabm/pyEpiabm/tests/test_unit/test_core/test_person.py
```diff
@@ -110,6 +110,17 @@ def test_vaccinate(self):
...
+    def test_set_id(self):
+        init_id = self.microcell.id + "." + "." + \
+                  str(len(self.microcell.persons) - 1)
+        self.assertEqual(self.person.id, init_id)
+        self.person.set_id("2.3.4.5")
+        self.assertEqual(self.person.id, "2.3.4.5")
+        self.assertRaises(TypeError, self.person.set_id, 2.0)
+        self.assertRaises(ValueError, self.person.set_id, "0.1")
+        self.assertRaises(ValueError, self.person.set_id, "1234")
+        self.assertRaises(ValueError, self.person.set_id, "0.0.0.0.5")
```
KCGallagher, line 122: These are great tests, but you should be consistent - in other tests you also check for errors with IDs that are too short (e.g. "0.0.0"), but you don't check for a series of numbers that are not period-separated. It would be good to have al these tests in all cases

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
```diff
@@ -217,8 +218,13 @@ def test_print_population(self, mock_copy, mock_read):
         actual_df = mock_copy.call_args.args[0]
         if version.parse(pd.__version__) >= version.parse("1.4.0"):
             expected_df = self.df.copy()
-            expected = expected_df.drop('household_number', axis=1)
-            actual = actual_df.drop('household_number', axis=1)
+
+            # The type of the id is converted from float to string, so we
+            # drop the cell and microcell columns
```
KCGallagher, line 223: Why is this? We should still check these match, and just convert back on the modified version. Better still, don't modify the data frame in the code, just convert to a string when you set the ID

KCGallagher (changes requested):
Great work! Sorry there are always a lot of comments on the first PR, but this is great quality work, and I've just added a lot of small corrections, as well as a few wider suggestions

pyEpiabm/pyEpiabm/core/cell.py
mghosh00: Yes, I will do this

pyEpiabm/pyEpiabm/routine/file_population_config.py
mghosh00: The id is currently a float in the csv file, so if we were to take it as a string, it would become "1.0" instead of "1" (which we want). We can look to see if there are ways to convert differently

pyEpiabm/pyEpiabm/routine/file_population_config.py
mghosh00: This can be changed, shouldn't be hard, but we'll still need to split the string

pyEpiabm/pyEpiabm/routine/file_population_config.py
mghosh00: Apologies, I meant line 117 not 115, but after the discussions we've had I will remove this comment now - this issue has been sorted

pyEpiabm/pyEpiabm/routine/file_population_config.py
mghosh00: Ah I see, yes will change these

pyEpiabm/pyEpiabm/tests/test_unit/test_core/test_person.py
mghosh00: Sure, will do

pyEpiabm/pyEpiabm/routine/file_population_config.py
KCGallagher: In that case perhaps we could just convert float -> int -> str when we feed it into the `set_id` function? But also I would be happy for you to change the output format in the csv files from `print_population` etc, as these don't need to be floats (or I can do this) - given the new form of the ID there is no reason not to be consistent here

pyEpiabm/pyEpiabm/routine/file_population_config.py
KCGallagher: You can do this in the list comp too:

```
microcell_ids = [microcell.id.split('.')[-1] for microcell in cell.microcells]
```

pyEpiabm/pyEpiabm/core/cell.py
mghosh00: Checked duplicates for microcell, household and person, but not sure how to do this for cell without adding an extra parameter to the set_id function containing a list of all other cells, which does not seem like a good idea

pyEpiabm/pyEpiabm/core/person.py
mghosh00: I don't quite understand this suggestion - the id will only be changed if it is of the correct format perhaps?

pyEpiabm/pyEpiabm/routine/file_population_config.py
mghosh00: I think it'd be best to stick with having a string input to the set_id function, and the output format in the csv files are currently strings I believe, as they have changed now. I don't quite understand - the id needs to be changed from a float to a string somewhere, so is it not ok to do it here?

pyEpiabm/pyEpiabm/tests/test_unit/test_routine/test_file_population_config.py
mghosh00: Ah yes, thanks, have made the change

pyEpiabm/pyEpiabm/core/cell.py
mghosh00: Actually, we are going to take the above approach (passing a list of cells to the set_id function) as this is done similarly in another function, but let us know if you would rather an alternative approach

pyEpiabm/pyEpiabm/routine/file_population_config.py
mghosh00: I actually ended up changing the approach to make it easier, by setting
`microcell_id_csv = cell_id_csv + "." + str(int(line["microcell"]))`
on line 90.

pyEpiabm/pyEpiabm/routine/file_population_config.py
KCGallagher: Ahh I see, I was thinking about the wrong csv! I think I am happy for this to stay as is then. I might have a slight preference for reading in the values as integers (as this is what they really are) and then just converting to string to append to the full id (with the cell etc). To do this, you set `dtype={"cell":int}` in the `read_csv` call (or even just `dtype=int` if all elements should be integers, which I think they should!

pyEpiabm/pyEpiabm/core/cell.py
KCGallagher: I see. Yes I don't love this - where else do we do something similar? I am almost tempted to store the population as an attribute of the cell (like the cell is an attribute of the microcell) to avoid this. But if you have seen it somewhere else then I suppose it is not a big deal, so happy to work with this

KCGallagher (approved):

pyEpiabm/pyEpiabm/routine/file_population_config.py
mghosh00: Ah ok sure, I will make this change

pyEpiabm/pyEpiabm/core/cell.py
abbie-evans: In cell.py, the function find_nearby_cells uses a list of cells

pyEpiabm/pyEpiabm/routine/file_population_config.py
```diff
@@ -95,14 +101,14 @@ def make_pop(input_file: str, random_seed: int = None, time: float = 0):
                 cell.set_location(location)
 
             # Raise error if microcell exists, then create new one
-            for microcell in cell.microcells:
-                if microcell.id == line["microcell"]:
-                    raise ValueError(f"Duplicate microcells {microcell.id}"
-                                     + f" in cell {cell.id}")
+            microcell_ids = [microcell.id for microcell in cell.microcells]
+            if microcell_id_csv in microcell_ids:
+                raise ValueError(f"Duplicate microcells {microcell_id_csv}"
```
KCGallagher, line 106: This doesn't read quite right once you substitute in the values in {} - maybe 'Duplicate microcells: {...} already exists in cell {...}'?

pyEpiabm/pyEpiabm/tests/test_func/test_closure_functional.py
```diff
@@ -54,7 +54,9 @@ def test_closure_present(self, mock_read, mock_csv):
             "test_input.csv", self.sim_params, self.file_params,
             HelperFunc.sweep_list_initialise())
 
-        mock_read.assert_called_with('test_input.csv')
+        mock_read.assert_called_with('test_input.csv',
```
KCGallagher, line 57: So in general any repetition of long(-ish) parameter chains like this is bad practice - if we were to chain one of these argument types later then we would have to go through and change all five appearances of that in this file. It also makes the code longer and less readable. Given the same arguments are repeated five times, it would be better to save this as a list in the setUp method (`self.read_params = [...]), and then just check mock_read was called with that list each time throughout the file. Does that make sense?

KCGallagher: Ditto in other files too - sadly I think no easy way to avoid duplication across files.

KCGallagher (changes requested):
Almost there!

pyEpiabm/pyEpiabm/intervention/travel_isolation.py
```diff
@@ -75,6 +75,15 @@ def __call__(self, time):
                                     selected_household = random.choice(
                                         existing_households)
                                     selected_household.add_person(person)
+
+                                    # Have to update the person's id to
+                                    # account for their new household
+                                    person.set_id(selected_household.id +
```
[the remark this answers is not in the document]
abbie-evans: Unfortunately without this line the test fails as it states that there are duplicate ids due to people moving household. We have discussed this with Lara, and she is going to look into it and discuss it on Wednesday

pyEpiabm/pyEpiabm/routine/file_population_config.py
```diff
@@ -83,9 +84,14 @@ def make_pop(input_file: str, random_seed: int = None, time: float = 0):
         # Store current cell
         current_cell = None
         # Iterate through lines (one per microcell)
-        for _, line in input.iterrows():
+        for line in input.itertuples():
+            line = line._asdict()
+            # Converting from float to string
+            cell_id_csv = str(line["cell"])
```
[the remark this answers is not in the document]
mghosh00: This is a method on named tuple, and though I agree this is normally bad practice, it seems that it is only underscored so that it does not clash with a potential field name. But if you'd prefer I can change it

pyEpiabm/pyEpiabm/tests/test_func/test_closure_functional.py
mghosh00: All sorted

pyEpiabm/pyEpiabm/routine/file_population_config.py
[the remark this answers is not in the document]
KCGallagher: I think the suggestions in the first comment would be better practice if they work, or do they throw errors? I haven't come across this issue before, but can investigate this locally. I can't find these references to `_fields` you mention - it doesn't seem to be on that line? https://github.com/SABS-R3-Epidemiology/epiabm/blob/7d8b8b08439fa2149e549eb5a46fb348324ce5c0/pyEpiabm/pyEpiabm/routine/file_population_config.py#L132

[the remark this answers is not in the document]
mghosh00: I just mean that if I keep it as a named tuple (and do not use _asdict()) then I would have to change line 132 to read
`if (('household_number' in line._fields) and`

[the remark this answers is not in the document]
mghosh00: So if I do or do not use _asdict(), I will still have at least one line in the code with a function/attribute being accessed with a leading underscore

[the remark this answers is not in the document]
KCGallagher: Hmm I know this is annoying but I do want to avoid using these private methods if we can. The pythonic alternative for L132 would be `if hasattr(line, 'household_number') and` - would you mind checking that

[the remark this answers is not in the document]
mghosh00: Done, I have made the changes

pyEpiabm/pyEpiabm/intervention/travel_isolation.py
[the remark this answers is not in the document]
mghosh00: I have changed add_household so that it accepts a new boolean parameter, which determines whether or not the id of the person is changed when add_household is called. I thought this would be a good idea as add_household is not only called in the travel isolation but is also called at different points when redistributing people between households, so it is useful to distinguish whether or not the ids of the people should be set


I wasn't sure about whether to set change_id to True or False in some places, so would be good to discuss at some point when exactly we want the id of the person to be set in all the different places

pyEpiabm/pyEpiabm/core/microcell.py
```diff
@@ -123,27 +141,41 @@ def add_place(self, n: int, loc: typing.Tuple[float, float],
...
+        change_id : bool
+            Boolean representing whether we wish to set the id of the people
+            of the household when the function is called or not
 
         """
         if len(people) != 0:
             household = Household(self, loc=self.location)
-            for person in people:
+            for i in range(len(people)):
+                person = people[i]
```
KCGallagher, line 160:
```suggestion
            for i, person in enumerate(people):
```

This is a really handy construction!

KCGallagher (changes requested):
Does the most recent change not negate the need for the `change_id` argument in `add_household`?

pyEpiabm/pyEpiabm/intervention/travel_isolation.py
```diff
@@ -75,6 +75,11 @@ def __call__(self, time):
                                     selected_household = random.choice(
                                         existing_households)
                                     selected_household.add_person(person)
+                                    logging.info(f"Person {person.id} has "
+                                                 f"finished isolating and "
+                                                 f"has moved to household "
```
KCGallagher, line 80: Does this work? I would normally expect to these strings explicitly concatenated using an addition operator (+). Also you don't need an f-string for the sections that don't have any curly brackets - these can just be normal string, with no f before the quotation marks!

pyEpiabm/pyEpiabm/intervention/travel_isolation.py
```diff
@@ -93,7 +98,7 @@ def __call__(self, time):
                                     person.household.persons.remove(person)
                                     # Put in new household
```
KCGallagher, line 99:
```suggestion
                                    # Move to temporary household
                                    # N.B Person ID is not changed
```

pyEpiabm/pyEpiabm/core/microcell.py
```diff
@@ -123,27 +141,40 @@ def add_place(self, n: int, loc: typing.Tuple[float, float],
             self.cell.places.append(p)
             self.places.append(p)
 
-    def add_household(self, people: list):
+    def add_household(self, people: list, override_person_id: bool = True):
```
KCGallagher, line 144: Sorry I have changed my mind - "update" would be a better verb than "override"!

pyEpiabm/pyEpiabm/tests/test_unit/test_core/test_microcell.py
```diff
@@ -70,14 +79,36 @@ def test_setup(self, n=5):
...
+                         original_household.id + ".0")
 
     @mock.patch('logging.info')
     def test_logging(self, mock_log):
-        self.microcell.add_household(self.microcell.persons)
+        # Check that the logger is called if there are no people in the
+        # household
+        self.microcell.add_household(self.microcell.persons,
+                                     override_person_id=True)
+        mock_log.assert_called_once()
```
KCGallagher, line 104: Given both tests check that the logging function is called, can we check that it is being called with the right message in each scenario? Should just be a string that matches the log message (with curly brackets in f-strings filled in by the relevant info)

pyEpiabm/pyEpiabm/tests/test_unit/test_intervention/test_travel_isolation.py
```diff
@@ -77,13 +79,15 @@ def test___call__(self, mock_random):
         self.assertIsNone(self.person_introduced.travel_isolation_start_time)
         self.assertEqual(len(self._microcell.households), 1)
         self.assertEqual(len(self.person_symp.household.persons), 3)
-
+        mock_log.assert_called()
```
KCGallagher, line 82: I really don't like this function. Can we check it was called the right number of times (and ideally with the right argument)

KCGallagher (changes requested):
Sorry for more changes - I promise there's nothing big here - some typing points and two quick tests, but no functional changes.

pyEpiabm/pyEpiabm/intervention/travel_isolation.py
mghosh00: The f strings have worked fine when I've tested them without explicit concatenation (they give the correct logging/error messages) and the f-string for sections without curly braces keep the format/alignment of the string the same. You can have multi-line f strings like this which concatenate without the plus operator and it works fine

pyEpiabm/pyEpiabm/tests/test_unit/test_core/test_microcell.py
mghosh00: Will do

pyEpiabm/pyEpiabm/tests/test_unit/test_intervention/test_travel_isolation.py
mghosh00: Will do

pyEpiabm/pyEpiabm/intervention/travel_isolation.py
KCGallagher: You learn something new everyday!

KCGallagher (approved):
Brilliant work @mghosh00 and @abbie-evans on this PR - thanks for your perseverance on this and all the hard work to get this finished! I am very happy with this addition, and you both adapted well to the issues coding this up that I had not anticipated, particularly regarding travel isolation.

PR #253, opened 2024-01-24
Waning immunity

### LINKED ISSUE #251: Bypassing Recovered compartment feature
We would like to begin implementing waning immunity for Epiabm. Currently once someone has Recovered, they remain there, and we would like to change this.

Initially, we will update the simulation so that once someone has Recovered, they immediately return to the Susceptible compartment. This will be done via the transition matrices and will only occur if the user passes a flag to the simulation. Further issues will cover implementation of waning immunity.

**Comparison to CovidSim**
Not currently implemented in CovidSim.

**Additional context**
This is the first step towards having waning immunity implemented in Epiabm.

tomcodewizard:
## Summary

We have implemented waning-immunity to epiabm, such that people in the Recovered compartment will move after a specified number of days to the Susceptible compartment.

We have added tests for our modifications.

## Further work

We now need to be able to change the transition probabilities between compartments, dependent on the waning immunity of a person. This will involve line fitting to data values found from the literature.

## Checklist

## Closing issues

Closes #251


pyEpiabm/pyEpiabm/core/person.py
```diff
@@ -258,3 +259,14 @@ def set_id(self, id: str):
             raise ValueError(f"Duplicate id: {id}.")
 
         self.id = id
+
+    def set_time_of_recovery(self, time: float):
+        """Records the time at which a person enters the Recovered compartment
```
KCGallagher, line 264:
```suggestion
        """Records the time at which a person enters the Recovered compartment.
```

python_examples/basic_infection_history_simulation/simple_parameters.json
```diff
@@ -334,6 +336,29 @@
...
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0
```
KCGallagher, line 360: These don't all need to be on separate rows, if they're all zero - I think it is much less readable.

KCGallagher: (Ditto for other json files)

KCGallagher (changes requested):
Overall great work! One thing this is missing is an update to the functional tests. There are different ways you could do this, but you want to check the basic qualitative results we expect are preserved (see current simulation function [tests](https://github.com/SABS-R3-Epidemiology/epiabm/blob/bef76876139b4178f29a45211ed27cd28c40cca1/pyEpiabm/pyEpiabm/tests/test_func/test_sim_functional.py#L127) for examples). For example, it would be good to have a test with complete infection (and no waning immunity) to check everyone goes to recovered, and then rerun with waning immunity to check that no longer happens. These simulations should be very simple (i.e. one cell and one household is fine) and be quick to run. These are just suggestions, but have a think and a look at the previous work for ideas, and let me know if you want to chat about it.

Can be a separate PR, but do we also want to include waning immunity from vaccination as well as infection?

KCGallagher (approved):

PR #267, opened 2024-06-20
Recording the serial interval

### LINKED ISSUE #262: Keeping track of secondary infections
We wish to record the number of people each person infects over the course of the epidemic (each time they are re-infected) so that we can find the R_t value at each time step. The dataframes produced will be further used in the sampling package EpiOS.

I will change the `Person` class to record a list of secondary infections each person has during the epidemic. I will also update the simulation to record and output a dataframe containing this information.

mghosh00:
## Summary

I have added the ability to record serial intervals of all infection events (which can be used to generate a distribution of serial intervals). The output of this is in a file named serial_intervals.csv.

## Checklist

## Closing issues

Closes #262


KCGallagher:
@mghosh00 just following up on this so the repo is up to date for publication - how much of this is actually intended to go onto main? For example I see the (untested) triple sweep work here, which I assume was just a simple debugging test and so will not actually be merged into the main repo. If this is the case, it could be more appropriate to close the PR, and perhaps delete the branch if we do not now need this code.

mghosh00:
Closes #273

pyEpiabm/pyEpiabm/sweep/place_sweep.py
```diff
@@ -57,6 +57,18 @@ def __call__(self, time: float):
...
+                            # to be the exposure period of the infectee
+                            inf_to_exposed = \
+                                (time - infector.infection_start_times[-1])
+                            infectee.set_exposure_period(inf_to_exposed)
+                            # Finally, store the infector's latency period
+                            # within the infectee, which will be used in
+                            # calculating the generation_time
+                            (infectee
+                             .set_infector_latent_period(infector
+                                                         .latent_period))
```
KCGallagher, line 71: I don't love that this code is repeated in each sweep, and would rather it was handled in the queue sweep (or an infection sweep) to avoid repetition. If this code has to be in this file (i.e. because the infector is only known in these sweeps, as I can't remember), then perhaps consider a separate function, that might be a method of the abstract sweep class, which each individual sweep class can call.

python_examples/basic_infection_history_simulation/simulation_outputs/output.csv
```diff
@@ -1,62 +1,62 @@
...

+12.0,99,0,0,0,0,0,0,0,1,0,0

+13.0,99,0,0,0,0,0,0,0,1,0,0

+14.0,99,0,0,0,0,0,0,0,1,0,0

+15.0,99,0,0,0,0,0,0,0,1,0,0

+16.0,99,0,0,0,0,0,0,0,1,0,0
```
KCGallagher, line 18: Did you mean to commit an updated version of this file? It seems to be a simulation where the pandemic doesn't take off and doesn't exemplify this software like the previous one. Might be best to revert these changes.

python_examples/basic_infection_history_simulation/triple_simulation_flow.py
```diff
@@ -0,0 +1,75 @@
+#
```
KCGallagher, line 1: Do you mean to commit this file to main? I don't think we necessarily need it there, as I am not sure that it is useful to a public audience (and was just for internal debugging). If it has wider use then we should document this in the readme.

python_examples/basic_infection_history_simulation/triple_simulation_outputs/output.csv
```diff
@@ -0,0 +1,22 @@
+time,InfectionStatus.Susceptible,InfectionStatus.Exposed,InfectionStatus.InfectASympt,InfectionStatus.InfectMild,InfectionStatus.InfectGP,InfectionStatus.InfectHosp,InfectionStatus.InfectICU,InfectionStatus.InfectICURecov,InfectionStatus.Recovered,InfectionStatus.Dead,InfectionStatus.Vaccinated
```
KCGallagher, line 1: Ditto to previous point - what do these simulation results tell us? I thought they were just for our debugging, but if they have wider interest then we need to document that

pyEpiabm/pyEpiabm/sweep/place_sweep.py
mghosh00: Ah yes, that makes sense, will work on that!

python_examples/basic_infection_history_simulation/simulation_outputs/output.csv
mghosh00: Thanks for catching this, and no this was a mistake! (same for all the files you mentioned below)

python_examples/basic_infection_history_simulation/triple_simulation_flow.py
mghosh00: See prev comment (mistake)

python_examples/basic_infection_history_simulation/triple_simulation_outputs/output.csv
mghosh00: See prev comment (mistake)

PR #274, opened 2024-12-05
Adding the Northern Ireland simulation

mghosh00:
## Summary

I have added the flow file and .json parameters file for the Northern Ireland simulation.


KCGallagher (approved):

KCGallagher:
Looks great - only comment is that it might be helpful to add a short description of this new simulation into the python examples [readme](https://github.com/SABS-R3-Epidemiology/epiabm/blob/secondary-infections/python_examples/README.md), so other users can easily find and understand it!

PR #275, opened 2025-04-19
Kc gallagher patch 1

KCGallagher:
Add logo image


KCGallagher:
@NicholasFan235 I am just updating the github workflows so they run - and found that your C++ workflows specify an older version of ubuntu to use (whereas all the python workflows just run on the latest ubuntu). This version is no longer supported by github workflows, and so those runs are automatically cancelled, however updating to the latest ubuntu version causes a different issue with the cmake. Would you be able to take a look at this?

PR #276, opened 2025-06-26
Age groups

laraherriott:
## Summary

Added a new function to update the age groups after manually assigning ages to children in households of three people or  more.

## Further work

NA

## Checklist

- [] Breaking change (fix or feature that would cause existing functionality to change)
- [] I have read the **CONTRIBUTING** document.
- [] My code follows the code style of this project.
- [] I have updated the wiki with new parameters/model functionality

### DIFF
--- a/pyEpiabm/pyEpiabm/core/person.py
+++ b/pyEpiabm/pyEpiabm/core/person.py
@@ -93,6 +93,7 @@ class Person:
         """Update the age_group attribute based on the current age value.
         This should be called after the age is set directly (e.g., in household
         age assignment) to ensure the age_group is consistent with the age.
+        
         """
         if Parameters.instance().use_ages and self.age is not None:
             self.age_group = (self.age // 5)
--- a/pyEpiabm/pyEpiabm/core/person.py
+++ b/pyEpiabm/pyEpiabm/core/person.py
@@ -89,6 +89,14 @@ class Person:
             # same age group (to conserve same output structure)
             self.age_group = 0
 
+    def update_age_group(self):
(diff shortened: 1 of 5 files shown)


pyEpiabm/pyEpiabm/core/person.py
```diff
@@ -89,6 +89,15 @@ def set_random_age(self, age_group=None):
...
             self.age_group = 0
 
+    def update_age_group(self):
+        """Update the age_group attribute based on the current age value.
+        This should be called after the age is set directly (e.g., in household
+        age assignment) to ensure the age_group is consistent with the age.
+
+        """
+        if Parameters.instance().use_ages and self.age is not None:
+            self.age_group = (self.age // 5)
```
KCGallagher: Does this match with age groups exactly? In particular I am worried that we have hard coded that the top age group is 80+, and so this would create too many age groups, that could result in indexing errors when trying to find age-group specific parameter values for people over 85

KCGallagher (changes requested):

pyEpiabm/pyEpiabm/core/person.py
laraherriott: The maximum age a person can have in the model currently is 84, so this wouldn't cause an issue currently. I agree it would be better to ensure there is a maximum age group for all those over 80 though in case of future changes, so have updated the function accordingly.

KCGallagher (approved):
All looks good! Though, just to check, have you plotted the age histograms by both age and age group to confirm that they match?

